PR #11, opened 2018-09-26
Add pyqubo to setup.py

kotarotanahashi:
Added pyqubo to setup.py

@randomir @arcondello It would be great If you could add `pyqubo` to ocean's dependencies.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,8 @@ install_requires = ['dwavebinarycsp>=0.0.2,<0.1.0',
                     'dwave-networkx>=0.6.1,<0.7.0',
                     'dwave-system>=0.5.0,<0.6.0',
                     'dwave-qbsolv>=0.2.7,<0.3.0',
-                    'dwave-neal>=0.4.0,<0.5.0'
+                    'dwave-neal>=0.4.0,<0.5.0',
+                    'pyqubo>=0.0.5'
                     ]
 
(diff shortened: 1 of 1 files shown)


arcondello:
Hi @kotarotanahashi , sorry to be so late in replying to you, we would really like to merge pyqubo into the sdk! One thing that is preventing us from merging at this time is the convention `Qubit` to describe a {0, 1} variable, which might be confusing to Ocean users. In Ocean we use the convention `Binary` for {0,1} and `Spin` for {-1,+1}.

kotarotanahashi:
@arcondello No problem! 
> the convention `Qubit` to describe a {0, 1} variable, which might be confusing to Ocean users
I can replace `Qubit` with `Binary`. Does that work? If you notice more problems, please let me know!

arcondello:
Yes I think that would do it, thanks!

kotarotanahashi:
@arcondello I replaced the class name "Qbit" to "Binary" not to confuse ocean users.
Thank you for pointing out the issue.
https://github.com/recruit-communications/pyqubo/releases/tag/0.2.0

arcondello:
Great! If you could just update the PR for the new version we'll merge. Or we can do it on our side

kotarotanahashi:
Hi, @arcondello , I updated the version of pyqubo, and resolved a conflict. It seems that we are ready to merge it.

PR #15, opened 2018-10-24
Platform-conditional dwavebinarycsp flavour install

randomir:


randomir:
Windows build failing because wheel is not available for `dwave-neal` for python 3.4@windows.

PR #21, opened 2018-12-24
Add dwave.samplers and dwave.composite namespace

arcondello:
Create a namespace for Samplers and Composites populated by entry points.

We could also consider having this as a separate package rather than part of the SDK.

### DIFF
--- a/dwave/samplers/__init__.py
+++ b/dwave/samplers/__init__.py
@@ -1,0 +1,20 @@
+# Copyright 2018 D-Wave Systems Inc.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#
+# ================================================================================================
+from pkg_resources import iter_entry_points
+
+# add the samplers to the module from entrypoints, name conflicts override
+globals().update((ep.name, ep.load()) for ep in iter_entry_points('dwave.samplers'))
+del iter_entry_points  # remove from namespace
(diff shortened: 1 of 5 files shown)


randomir (approved):
LGTM! :)

setup.py
```diff
@@ -27,6 +29,8 @@
 else:
     exec(open("./dwaveoceansdk/package_info.py").read())
 
+os.chdir(os.path.dirname(os.path.abspath(__file__)))
```
randomir: Not sure why do you need this?

dwave/samplers/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+    setup(
+        entry_points=entry_points,
+        ...
+    )
+
+"""
+from pkg_resources import iter_entry_points
+
+# add the samplers to the module from entrypoints, name conflicts override
+globals().update((ep.name, ep.load()) for ep in iter_entry_points('dwave.samplers'))
```
randomir: Oh, since now we have samplers in Hybrid as well, maybe we want to differentiate those from dimod samplers?

randomir: Something like:
```
dwave.dimod.samplers
dwave.hybrid.samplers
```
Or think of a way to reconciliate the two. Same goes for `dwave.composites`.

setup.py
arcondello: No idea, probably don't

dwave/samplers/__init__.py
arcondello: how about
```
dwave.samplers
dwave.composites
dwave.runnables  # this could be further subdivided
```

randomir: I don't love it, but it could work.

On a somewhat related note, I'm getting fond of `from dwave import dimod, hybrid`.

arcondello: Are you proposing making dimod and hybrid into namespace packages? That would be easy to do, but at some point it would make sense to just make them all into a monolith package.

Or do you mean like the above where just the samplers/composites are present? I think it would be extremely confusing for `from dwave import hybrid` to contain different contents/structure than `import hybrid`.

randomir: The former. I'm toying with the idea.

arcondello:
Closing this as (extremely) stale

PR #34, opened 2019-08-21
Add packages as submodules

arcondello:
This PR does a couple things:
* Merges the [docs](https://github.com/dwavesystems/docs) directly into the sdk - this will require re-configuring the readthedocs
    * Serves as an alternative to https://github.com/dwavesystems/docs/pull/62
* Adds the dependencies of the sdk as submodules - this requires pinning the version in the sdk

Other considerations:
* To get all of the docs together, this will require adding projects into the sdk that were not previously included in the requirements directly, e.g. [dwave-cloud-client](https://github.com/dwavesystems/dwave-cloud-client)
* In my opinion this complicates the repo significantly and users who might want to edit the sdk-level documentation might bounce off it
* I am not sure how git's symlink support translates to windows - though https://stackoverflow.com/a/52097145/8766655 implies that newer versions of git will support it


arcondello:
As of the first commit only dimod as been added (to show how it could be done). Also the doctests are failing both for syntax errors and because the QPU is not available to circleci.

arcondello:
Merged into dev for now so that the changes are not overwhelming

PR #37, opened 2019-12-04
Add the meta-docs to the sdk

arcondello:
Contains some of the commits from #34 but contains only the docs to make it cleaner. Subsequent PRs will add the Ocean packages as submodules.

Note that the docs are currently failing tests and need some fixes.


arcondello:
I will fix the doc examples in a subsequent PR so that it's easier for @JoelPasvolsky to check the changes.

PR #38, opened 2019-12-04
Sync OS/Python versions across Ocean wheels

randomir:

PR #40, opened 2019-12-07
Add all Ocean packages as submodules and pin version numbers

arcondello:
All of the docs build, but there are some issues:
* Some packages to not have images showing correctly, this can be resolved by fixing the image paths and then adding all of the various `_static` directories to the main `conf.py`.
* Packages with C documentation/components do not have those components showing correctly. This can be fixed an an ad hoc way, but a more general solution would be nice.
* A lot of the cross links in the meta-docs were broken, specifically the ones that invoke other package's `index` (e.g. `dimod:index`). This is broken because I have made new `shortindex.rst` files that include only the information relevant for the unified docs, but not all the other links to LEAP etc. We might want to rename then and replace all of the original `index.rst` files with `longindex.rst` or similar.
* Some of the `.. include::` directives (especially for the licenses) are broken. This can be resolved by making sure `.. include::` directives only refer to things in the `docs/` sub-directory.

I will make fixes in subsequent PRs.


arcondello:
The docs are building correctly (you can see them in the circle-ci artefacts), but the doctests are still failing (as in #37).

Also, the osx and py3.4 tests will be fixed by #38.

arcondello:
I will make fixes in subsequent PRs.

PR #46, opened 2020-01-15
Unified docs: replace dir copies with symlinks

JoelPasvolsky:
@arcondello, would you please update submodules for this step, so I can clear some of the build warnings, I'll work on fixing the rest on a cleaner build.


JoelPasvolsky:
These failures are penaltymodel version mismatches, not related to my changes

arcondello:
Do we want to merge this simultaneously with the new tagged versions of the various projects? Otherwise won't the docs be broken?

JoelPasvolsky:
Yes

arcondello:
Was that a response to the first or second question?

JoelPasvolsky:
Okay, I updated the submodules

PR #47, opened 2020-01-17
Fix metadoc refs to submodules

JoelPasvolsky:
This fixes the broken metadoc links. It also stops the build warnings for missing index includes on README and LICENSE, but it's an inelegant workaround that I hope to replace with something better later.


JoelPasvolsky:
The failed checks are the usual penaltymodel problem. For the build warnings I'll experiment later with using the ifconfig:: or only:: directive to replace the dummy files.

JoelPasvolsky:
I ran some quick tests and I'm pretty sure I'll be able to replace the dummy files with a more elegant solution. It will require updates to the submodule repos so it's worthwhile to merge this in the meanwhile to get the SDK docs working.

PR #51, opened 2020-02-08
Lower bound sphinx version in docs/requirements.txt

### LINKED ISSUE #43: Sphinx must be version >=1.4 for sphinx.ext.githubpages
**Description**
requirements file has just "sphinx" but sphinx.ext.githubpages is supported only from 1.4. 

**To Reproduce**
run with older version and get an error

**Expected behavior**
requirement should be sphinx>=1.4
**Environment:**
 - OS: Ubuntu 16.04.4 
 - Python version:  3.5

**Additional context**

arcondello:
Closes #43

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,5 +1,5 @@
 dwave-ocean-sdk
-sphinx
+sphinx>=2.0.0
 sphinx_rtd_theme
 recommonmark
 


docs/requirements.txt
```diff
@@ -1,5 +1,5 @@
 dwave-ocean-sdk
-sphinx
+sphinx>=2.0.0
```
JoelPasvolsky: We currently have for https://github.com/dwavesystems/minorminer/blob/master/docs/requirements.txt a requirement that sphinx==1.*
I think that was needed for the Python IF to build

JoelPasvolsky: Note that https://github.com/dwavesystems/minorminer/issues/102 is now closed. @arcondello, any reason not to proceed with this merge?

randomir:
We lower-bound sphinx to 4.0 since #128.

PR #59, opened 2020-02-24
Leap2

arcondello:


arcondello:
Now the doctest failures are just normal doctest failures, not build issues. So merging

PR #62, opened 2020-03-04
Add Ocean concepts

JoelPasvolsky:
To better integrate the unified Ocean docs, this PR proposes a Concepts section. It also makes an additional couple of changes:
* Short descriptions of the packages on front page
* Responsive wide tables (instead of long scroll bars)

You can see the rendered docs here: https://sdk-docs.readthedocs.io/en/latest/

This is still a WIP, the concepts are not yet done, but let me know what you think of the proposal. 

Also, the SDK can pull in the content from the submodules but not vice-versa so without duplicating content, which I do not want to happen, either the intro sections in the submodules should just reference the content in Concepts or the Concepts would include the (reworked) content from the submodule repos: do we want these descriptions in the tools repos or should should be more Reference docs with the SDK doing all the introduction?


docs/concepts/bqm.rst
```diff
@@ -0,0 +1,52 @@
+.. _bqm_sdk:
+
+=======================
+Binary Quadratic Models 
+=======================
+
+The binary quadratic model (BQM) class contains
+Ising and quadratic unconstrained binary optimization (QUBO) models
+used by samplers such as the D-Wave system.
```
arcondello: I think we should also put the BQM equation here

docs/concepts/bqm.rst
```diff
@@ -0,0 +1,52 @@
...
+as an upper-diagonal matrix :math:`Q`, where diagonal terms are the linear coefficients
+and the nonzero off-diagonal terms the quadratic coefficients.
+
+.. math::
+
+    \text{QUBO:} \qquad E(\bf{x}| \bf{Q})  
+    =  \sum_{i\le j}^N x_i Q_{i,j} x_j
+    \qquad\qquad x_i\in \{0,1\}
+
+The :class:`.BinaryQuadraticModel` class can contain both these models and its methods provide
```
arcondello: contain -> encode

docs/concepts/bqm.rst
```diff
@@ -0,0 +1,52 @@
...
+
+   Introduces key concepts such as objective functions, Ising model, QUBOs, and graphs, explains
+   how these models are used to represent problems, and provides some simple examples.
+* :std:doc:`D-Wave Problem-Solving Handbook <sysdocs_gettingstarted:doc_handbook>`
+
+   Provides a variety of techniques for, and examples of, reformulating problems as BQMs.
+* :std:doc:`Solving Problems on a D-Wave System <oceandocs:overview/solving_problems>`
+
+   Describes and demonstrates the use of BQM in the context of Ocean software.
+
```
arcondello: Maybe a link to the new bqm section of the dimod docs?

Also should this whole section "live" in dimod and be imported rather than living in the sdk?

docs/concepts/embedding.rst
```diff
@@ -0,0 +1,14 @@
+.. _embedding_sdk:
+
+===============
+Minor-Embedding 
+===============
+
+To solve an arbitrarily posed binary quadratic problem directly on a D-Wave system requires mapping,
+called *minor embedding*, to a Chimera graph that represents the system's quantum processing unit.
```
arcondello: Link to glossary?

docs/concepts/samplers.rst
```diff
@@ -0,0 +1,59 @@
...
+Composites
+==========
+
+Samplers can be composed. The `composite pattern <https://en.wikipedia.org/wiki/Composite_pattern>`_
+allows layers of pre- and post-processing to be applied to binary quadratic programs without needing
+to change the underlying sampler implementation. We refer to these layers as `composites`.
+A composed sampler includes at least one sampler and possibly many composites.
+
+Examples of composites are :class:`~dwave.system.composites.EmbeddingComposite()`,
+used in the example above, and :class:`~dwave.system.composites.VirtualGraphComposite()`,
```
arcondello: I think we should not emphasize the VGC because of its QPU usage, I think RoofDualityComposite would be a good substitute.

docs/concepts/bqm.rst
JoelPasvolsky: This is what I wanted everyone's input on: I'm already tending to see the SDK as where most users go, and so am inclined to concentrate everything there, where it can most easily integrate. I see the submodules' repos' RTDs as used mostly by advanced users who will use maybe reference documentation but not introductory content. The disadvantage is that the content is further from the code source.

arcondello: I think either is fine. Especially for this sort of thing that is not code specific. As long as we have a reliable sphinx index scheme, I would be fine with just linking it from the relevant documentation. E.g. [asarray](https://docs.scipy.org/doc/numpy/reference/generated/numpy.asarray.html) doesn't spend any time talking about matrices or arrays.

docs/concepts/samplers.rst
JoelPasvolsky: What about a dimod one: SpinReversalTransformComposite? Might be better to promote that one for less experienced users, no?

arcondello: The `RoofDualityComposite` one is in dimod. The problem with SRTC is that it's already a server-side feature so it's somewhat redundant. I like RDC because almost no one is using it but most people should. I also think it's no more "advanced" than the VGC.

JoelPasvolsky: image

docs/concepts/samplers.rst
```diff
@@ -0,0 +1,61 @@
+.. _samplers_sdk:
+
+=======================
+Samplers and Composites 
+=======================
+
```
m3ller: Have you considered putting links at the top (i.e. the line here), so that users can just jump to the different subsections of this page easily?

docs/concepts/topology.rst
```diff
@@ -0,0 +1,83 @@
...
+
+============
+QPU Topology 
+============
+
+To solve a :term:`bqm` on the D-Wave system, you
+must map it to a graph that represents the topology of the system's
+qubits. For D-Wave 2000Q systems, this is the *chimera* topology; for 
+next-generation Advantage systems, this is the *Pegasus* topology.
+
```
m3ller: same suggestion with the subsection links here.

docs/concepts/bqm.rst
JoelPasvolsky: image

docs/concepts/samplers.rst
JoelPasvolsky: They do have: 
```
.. _samplers_sdk:

=======================
Samplers and Composites 
=======================
```
Do you mean something else?

docs/concepts/topology.rst
JoelPasvolsky: The headers in the RST file are rendered in the RTD theme as lines in the navigation bar:
image

If from another place we'll want to jump directly to a subsection here, we can either add a ref name above it then or use :doc:` <oceandocs:concepts/topology#subsection>`.

I'm not sure we'll want to jump to a subsection here and already now build generate hundreds of duplicate target warnings for the SDK because submodule repos have identical ref names.

m3ller: Oh I see. Thanks for letting me know. In that case, please ignore my suggestion.

docs/concepts/samplers.rst
m3ller: I had meant links to `Samplers`, `Hybrid Quantum-Classical Samplers` etc. But as you had already pointed out, these exist on the side panel. So please ignore my earlier suggestion.

docs/concepts/index.rst
```diff
@@ -0,0 +1,263 @@
...
+         A mathematical expression of the energy of a system as a function of
+         binary variables representing the qubits.
+
+      Pegasus
+         The D-Wave :term:`QPU` is a lattice of interconnected qubits. While some qubits
+         connect to others via couplers, the D-Wave QPU is not fully connected.
+         Instead, the qubits interconnect in an architecture known as Pegasus.
+         See a fuller description under :doc:`QPU Topology </concepts/topology>`.
+
+      Penalty function
```
arcondello: Might as well add `penalty model` as well, you'll need it later.

docs/concepts/index.rst
```diff
@@ -0,0 +1,263 @@
...
+         An algorithm for solving constrained optimization problems. In the context
+         of Ocean tools, penalty functions are typically employed to increase the energy
+         level of a problem’s :term:`objective function` by penalizing non-valid configurations.
+         See `Penalty method on Wikipedia <https://en.wikipedia.org/wiki/Penalty_method>`_
+
+      QPU
+         Quantum processing unit
+
+      QUBO
+         Quadratic unconstrained binary optimization.
```
arcondello: Why not also link wikipedia? Can be applied to others

docs/concepts/index.rst
```diff
@@ -0,0 +1,263 @@
...
+         of samples, in order of increasing energy.
+
+      SAPI
+         Solver API used by clients to communicate with a :term:`solver`.
+
+      Solver
+         A resource that runs a problem. Some solvers interface to the :term:`QPU`;
+         others leverage CPU and GPU resources.
+
+      Source
```
arcondello: Maybe `Source Graph` and `Target Graph`, source and target are used more generally in some places in Ocean I think

docs/concepts/index.rst
```diff
@@ -0,0 +1,263 @@
...
+      Embedding
+      Minor-embed
+      Minor-embedding
+         The nodes and edges on the graph that represents an objective function
+         translate to the qubits and couplers in :term:`Chimera`. Each logical qubit, in
+         the graph of the :term:`objective function`, may be represented by one or more
+         physical qubits. The process of mapping the logical qubits to physical
+         qubits is known as minor embedding.
+         See a fuller description under :doc:`Minor-Embedding </concepts/embedding>`.
+
```
arcondello: `Excited State` would also be useful to be able to link to

arcondello: `Ground state` as well

docs/concepts/solutions.rst
```diff
@@ -0,0 +1,91 @@
+.. _solutions_sdk:
+
+=========
+Solutions
```
arcondello: I think this should be `Samples`; if my problem is a CSP and the sampler returns an excited state then calling it a `solution` is pretty misleading

JoelPasvolsky: At the concept level I especially used "solutions" for new users that won't know what "samples" mean rather than try to be precise.

docs/concepts/index.rst
JoelPasvolsky:
```
Source
Source graph
```
And same for target

docs/concepts/index.rst
JoelPasvolsky: Added

docs/concepts/index.rst
JoelPasvolsky: Added both

docs/concepts/index.rst
JoelPasvolsky: Linked

PR #64, opened 2020-03-18
Fix most failures in doctest

JoelPasvolsky:
Fixes most doctest failures. This is a temporary solution -- yesterday I got a mock DWaveSampler working inside local doctests, more or less, so expect a next PR to expand the coverage of doctests that is reduced by this PR.

### DIFF
--- a/docs/examples/and.rst
+++ b/docs/examples/and.rst
@@ -150,12 +150,12 @@ The next code sets up a D-Wave system as the sampler.
       You can see this information by running :code:`dwave config inspect` in your terminal.
 
 >>> from dwave.system import DWaveSampler, EmbeddingComposite
->>> sampler = DWaveSampler(endpoint='https://URL_to_my_D-Wave_system/', token='ABC-123456789012345678901234567890', solver='My_D-Wave_Solver')
->>> sampler_embedded = EmbeddingComposite(sampler)
+>>> sampler = DWaveSampler(endpoint='https://URL_to_my_D-Wave_system/', token='ABC-123456789012345678901234567890', solver='My_D-Wave_Solver')   # doctest: +SKIP
+>>> sampler_embedded = EmbeddingComposite(sampler)       # doctest: +SKIP
 
 As before, we ask for 5000 samples.
 
->>> response = sampler_embedded.sample_qubo(Q, num_reads=5000)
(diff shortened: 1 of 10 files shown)


docs/examples/and.rst
```diff
@@ -220,13 +220,13 @@ the problem. Its last line prints a confirmation that indeed the two selected qu
...
 >>> from dwave.system.composites import FixedEmbeddingComposite
->>> sampler_embedded = FixedEmbeddingComposite(sampler, {'x': [0], 'z': [4]})
+>>> sampler_embedded = FixedEmbeddingComposite(sampler, {'x': [0], 'z': [4]})   # doctest: +SKIP
 >>> print(sampler_embedded.adjacency)     # doctest: +SKIP
 {'x': {'z'}, 'z': {'x'}}
 
 As before, we ask for 5000 samples.
 
->>> response = sampler_embedded.sample_qubo(Q_not, num_reads=5000)
+>>> response = sampler_embedded.sample_qubo(Q_not, num_reads=5000)      # doctest: +SKIP
```
arcondello: I think we should start the process of switching these over to `sampleset`

docs/examples/and.rst
```diff
@@ -220,13 +220,13 @@ the problem. Its last line prints a confirmation that indeed the two selected qu
...
 >>> for datum in response.data(['sample', 'energy', 'num_occurrences']):   # doctest: +SKIP
```
arcondello: Also think we should start using the `print(sampletset)` output

docs/examples/and.rst
JoelPasvolsky: I agree but not in this PR. I'll be going through all examples as part of my next PR to use a mock sampler in doctests and that will be the right time to update the examples, as I'll be doing that anyway to extend the doctest coverage by adjusting outputs to reliably reproducible ones

docs/examples/and.rst
JoelPasvolsky: Yes

JoelPasvolsky:
Submodules still to be updated: 
* https://github.com/dwavesystems/dwave_networkx/pull/161
* https://github.com/dwavesystems/dwave-neal/pull/76
* https://github.com/dwavesystems/dwave-cloud-client/pull/372
* https://github.com/dwavesystems/dwave-hybrid/pull/222
* https://github.com/dwavesystems/dwavebinarycsp/pull/100  -- MERGED

PR #75, opened 2020-05-29
Define BQP with links in BQM description

arcondello:


docs/concepts/bqm.rst
```diff
@@ -47,6 +47,13 @@ The :class:`.BinaryQuadraticModel` class can contain both these models and its m
 convenient utilities for working with, and interworking between, the two representations
 of a problem.
 
+Binary quadratic models are sometimes referred to as (unconstrained) binary
+`quadratic programs`_. We use BQM to avoid confusion with the
+`bounded error quantum polynomial time`_ (BQP) complexity class.
```
JoelPasvolsky, line 52: I wouldn't add the "We use BQM to avoid confusion with..." here

JoelPasvolsky (approved):

docs/concepts/bqm.rst
JoelPasvolsky: I think the first sentence, that states that BQMs can be thought of as binary quadratic programs, coming after a description of what BQMs are, does the work. I don't see how anyone at this point would confuse BQM and BQP. 
Someone somewhere else might be confused but not someone whose just read the text above, so at this point introducing BQP would make sense if you wanted to talk about BQP in relation to QA, but not to dismiss a confusion that is unlikely relevant here.

arcondello: How about
```
Binary quadratic models are sometimes referred to as (unconstrained) binary `quadratic programs`_. Not to be confused with BQP_, a quantum complexity class.
```

Another approach would be a wikipedia style
```
For the quantum complexity class, see BQP_
```
and put it at the top. [Example](https://en.wikipedia.org/wiki/Windsor_West_(provincial_electoral_district)) found using wikipedia's "random article" feature.

JoelPasvolsky: Yes, that would be good: For the quantum complexity class, see `bounded-error quantum polynomial time (BQP) <https://en.wikipedia.org/wiki/BQP>`_.

If you think many users coming to BQM have this confusion, it's worth putting at the top. My guess is that those are a small minority versus beginners looking for a definition of BQM, so would be better at the end.

arcondello:
Closing as (very) stale

PR #88, opened 2020-09-08
Ensure Getting Started examples are compliant with Pegasus

JoelPasvolsky:
Also updated references to Ocean classes to exploit integration with tools' repos.


docs/examples/hybrid_solver_service.rst
```diff
@@ -60,7 +62,7 @@ Formulate the Problem
 =====================
 
 For a social graph, `G`, this example simply builds a random sparse graph---using the
-`NetworkX <https://networkx.github.io/>`_ `random_geometric_graph()` function, which
+`NetworkX <https://networkx.github.io/>`_ :code:`random_geometric_graph()` function, which
```
arcondello, line 65: why not `:func:`random_geomatric_graph`?

JoelPasvolsky: Using `:func:`random_geomatric_graph`` results in what looks like a link but is not:
image

I can link it using `:func:`~networkx.generators.geometric.random_geometric_graph()``. I've made the change. It has the disadvantage that if NetworkX restructures, it breaks.

arcondello: I _think_ ```:func:`~networkx.random_geometric_graph` ``` should work

JoelPasvolsky: Unfortunately not

randomir: I'd encourage linking with function docs even for 3rd party packages. We mention NetworkX functions below in a few places as well.

docs/examples/map_kerberos.rst
```diff
@@ -84,10 +86,12 @@ Ocean's :doc:`dwave_networkx </docs_dnx/sdk_index>` can return a
 `minimum vertex coloring <https://en.wikipedia.org/wiki/Graph_coloring>`_ for a graph,
 which assigns a color to the vertices of a graph in a way that no adjacent vertices
 have the same color, using the minimum number of colors. Given a graph representing a
-map and a :term:`sampler`, the `min_vertex_coloring` function tries to solve the
+map and a :term:`sampler`, the :code:`min_vertex_coloring` function tries to solve the
```
randomir, line 89: Instead of just formatting as code, I would link with dwave-networkx docs.

randomir (approved):
LGTM.

JoelPasvolsky:
I think I've updated for all GS examples to directly link

PR #90, opened 2020-09-09
Update C++ section of contributing guide and include .clang-format

shpface:


CONTRIBUTING.rst
```diff
@@ -134,34 +134,23 @@ Documentation
...
+C++ code should be compatible with standard C++11.
 
-    if (condition) {
-        foo();
-    } else {
-        bar();
-    }
+Naming
+~~~~~~
+* File names should be lowercase with underscores and end with ``.cpp`` or ``.hpp``.
```
randomir, line 144: Google style guide allows dashes as well.

CONTRIBUTING.rst
```diff
@@ -134,34 +134,23 @@ Documentation
...
+~~~~~~
 
-  .. code-block:: c++
-
-    if (condition) {
-      foo();
-    } else {
-      bar();
-    }
+When starting a new C++ project, use clang-format with the .clang-format file included here.
```
randomir, line 153: Worth pointing out we base our style on [Google's](https://google.github.io/styleguide/cppguide.html). Also linking to it.

CONTRIBUTING.rst
```diff
@@ -134,34 +134,23 @@ Documentation
...
 
 Additional Style
```
randomir, line 155: Not sure we need this.

.clang-format
```diff
@@ -0,0 +1,14 @@
...
+
+ColumnLimit: 120
+NamespaceIndentation: Inner
+
+# Scaled by a factor of 2 such that the base indent is 4
+AccessModifierOffset: -2
+ConstructorInitializerIndentWidth: 8
+ContinuationIndentWidth: 8
+IndentWidth: 4
+ObjCBlockIndentWidth: 4
```
randomir, line 13: `ObjC`?

.clang-format
```diff
@@ -0,0 +1,14 @@
...
+Language: Cpp
+BasedOnStyle: Google
+
+ColumnLimit: 120
+NamespaceIndentation: Inner
+
+# Scaled by a factor of 2 such that the base indent is 4
+AccessModifierOffset: -2
+ConstructorInitializerIndentWidth: 8
+ContinuationIndentWidth: 8
```
randomir, line 11: Why deviate from Google's style for these?

randomir (commented):
It would be nice to provide a short code snippet that demonstrates the flavour of our naming and formatting guidelines. Not sure we can capture the essence in ~15 lines, though.

CONTRIBUTING.rst
```diff
@@ -134,57 +134,55 @@ Documentation
...
+* Variable names should be lowercase with underscores. Private data members should have a trailing underscore.
+* Global variable names should be ``ALL_CAPS_WITH_UNDERSCORES``.
+* Macros should be ``ALL_CAPS_WITH_UNDERSCORES``.
 
-Favor the use of the optional braces for single-line control statements, which enhance consistency and extensibility.
+Format
+~~~~~~
 
-Example:
+* When starting a new C++ project, use clang-format with the .clang-format file included here.
```
JoelPasvolsky, line 157: Nice to have:  "... with the `.clang-format <.clang-format>`_ file included ..."

CONTRIBUTING.rst
```diff
@@ -134,57 +134,55 @@ Documentation
...
-      bar();
-    }
+Naming
+~~~~~~
 
-Additional Style
-~~~~~~~~~~~~~~~~
+* File names should be lowercase with underscores or dashes and end with ``.cpp`` or ``.hpp``.
+* Namespaces should be lowercase with underscores.
+* Class names should be PascalCase. i.e. :code:`AdjArrayBQM`.
```
JoelPasvolsky, line 147: I feel bad but: "PascalCase; i.e., " (lowercase i after a period).
Same for "standard library; i.e.," and "underscores; i.e., " below

CONTRIBUTING.rst
```diff
@@ -134,57 +134,55 @@ Documentation
...
+~~~~~~
 
-Example:
+* When starting a new C++ project, use clang-format with the .clang-format file included here.
+* Our format is based on `Google C++ style guide <https://google.github.io/styleguide/cppguide.html>`_ with some exceptions:
 
-Use the following format
+  - The naming conventions are as stated above.
+  - Column width is limited to 120 characters.
+  - Indent widths are doubled so the base indent level is 4 spaces, line continuations indent 8 spaces, and access modifiers indent 2 spaces.
```
JoelPasvolsky, line 162: We're not doing 80 char columns?

.clang-format
```diff
@@ -0,0 +1,13 @@
+---
+Language: Cpp
+BasedOnStyle: Google
+
+ColumnLimit: 120
```
arcondello, line 5: We've more-or-less gone back to 80 in our python code, should this still be 120?

JoelPasvolsky (approved):

CONTRIBUTING.rst
arcondello: jinx :laughing:

.clang-format
randomir: I'm fine with 80, but then again I haven't written c++ in a long time..

arcondello: Ok, given that I think most folks will implement this as a commit hook or similar, I think it makes sense to keep it at 120 to allow users to occasionally exceed it (as we do in python). But I'll just add in writing that keeping it to 80 is preferred.

.clang-format
arcondello: I think it's "Scaled by a factor of 2 such that the base indent is 4", not sure why that's important though.

arcondello: @boothby , do you have insight? I know you use the same format [in minorminer](https://github.com/dwavesystems/minorminer/blob/main/.clang-format).

arcondello: If there is no specific reason, I would probably prefer to default to Google's style.

boothby: Huh.  I'm not actually sure.  Brad set that up and I've never changed it.

randomir (approved):
LGTM

PR #98, opened 2020-09-28
fix PyQUBO version

kotarotanahashi:
I will release a new C++ version of PyQUBO soon.
Before then, I would like to fix the version of PyQUBO to avoid unexpected error which might occur when installing a new PyQUBO.
After I will release a new PyQUBO, I will update the version of PyQUBO specified in dwave-ocean-sdk.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,7 @@ install_requires = [
     'penaltymodel-cache==0.4.1',
     'penaltymodel-lp==0.1.3',
     'penaltymodel==0.16.3',
-    'pyqubo>=0.4.0',
+    'pyqubo==0.4.0',
 ]
 
 # note: when updating the version of maxgap, it also must be updated in


randomir (approved):

randomir:
@kotarotanahashi, sounds good. How soon do you need a new version of Ocean that includes this patch?

kotarotanahashi:
@randomir Thank you for merging this branch soon. I am planning to release a new PyQUBO this week. I'd appreciate it if you could include this patch sooner.

randomir:
@kotarotanahashi, sure, I can do it right away, so we're not blocking you. We prefer having all versions pinned anyway.

kotarotanahashi:
@randomir Thank you.
>We prefer having all versions pinned anyway.
Sure. I will pin the version from the next time.

PR #101, opened 2020-10-19
Update pyqubo version to 1.0.6

kotarotanahashi:
- Update PyQUBO version to 1.0.6
- In the new version, the backend of PyQUBO is implemented in C++, and the execution time is 1000 times faster than the previous version.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,7 @@ install_requires = [
     'penaltymodel-cache==0.4.1',
     'penaltymodel-lp==0.1.3',
     'penaltymodel==0.16.3',
-    'pyqubo==1.0.6',
+    'pyqubo==1.0.7',
 ]
 
 # note: when updating the version of maxgap, it also must be updated in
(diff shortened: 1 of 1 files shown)


arcondello:
Hi @kotarotanahashi , are there other backwards compatibility breaks we should know about?

kotarotanahashi:
Hi @arcondello.
I change the interface of some methods in PyQUBO. Do you need the list of changes?
Compatibility with other ocean packages hasn’t changed this time.

kotarotanahashi:
@arcondello 
> are there other backwards compatibility breaks we should know about?

Main changes made in this version is described in the following release page.
https://github.com/recruit-communications/pyqubo/releases/tag/1.0.0

arcondello:
Hi @kotarotanahashi , thanks for the update. We were trying to determine if this change should increment the major version of our sdk. I talked with @randomir about it, and I think it is OK for this to just increment the minor version. Thanks for the PR!

kotarotanahashi:
@arcondello 
> I think it is OK for this to just increment the minor version

I see! Thank you for the review.

PR #103, opened 2020-10-26
Update .clang-format

arcondello:
* Change ColumnLimit to 80, otherwise it will try to use as much space as possible
* Change AccessModifierOffset to -3 to make it consistent with Google style

See also #90

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -2,10 +2,10 @@
 Language: Cpp
 BasedOnStyle: Google
 
-ColumnLimit: 120
+ColumnLimit: 80
 NamespaceIndentation: Inner
 
 # Scaled by a factor of 2 such that the base indent is 4
-AccessModifierOffset: -2
+AccessModifierOffset: -3
 IndentWidth: 4
 ...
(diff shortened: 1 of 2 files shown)


.clang-format
```diff
@@ -2,10 +2,10 @@
 Language: Cpp
 BasedOnStyle: Google
 
-ColumnLimit: 120
+ColumnLimit: 80
 NamespaceIndentation: Inner
 
 # Scaled by a factor of 2 such that the base indent is 4
-AccessModifierOffset: -2
+AccessModifierOffset: -3
```
randomir, line 9: Is Google's style "indent by one", or "indent by half indent width"?

I have always interpreted it as the latter, but have never checked.

randomir (commented):
80 char columns seem OK (especially after reading [this](https://stackoverflow.com/questions/33656800/clang-format-line-breaks)). Not sure about the access modifier offset.

.clang-format
arcondello: https://google.github.io/styleguide/cppguide.html#Class_Format says "indent by 1", though of course they also say "indent by 2". FWIW, https://github.com/google/styleguide complains about private/public/etc being 2, but not about the overall indent being 4.

arcondello:
Closing as (very) stale.

PR #104, opened 2020-11-06
Minor documentation updates

mcfarljm:
* Fix label in Ocean samplers table: `dwave-neal` -> `dwave-greedy`
* Update some node adjacency values for the Advantage system in the AND gate example.  Please review changes.  The previous text showed only a single node in the adjacency printout, and then it also used the same code for the `FixedEmbeddingComposite` example with Advantage as with 2000Q, which did not seem to be correct.  I got the new adjacency values using the "Advantage_system1.1" solver.
* 24652f9 and af94286 are both changes to link directly to topical subsections in the documentation that are being referenced in the text.  As a first-time reader, this would make it easier for me to find the information that is being referred to.  Happy to rebase and back these out though if this is not the preferred style.

### DIFF
--- a/docs/concepts/embedding.rst
+++ b/docs/concepts/embedding.rst
@@ -90,6 +90,8 @@ intentionally sets the embedding shown below to represent this same :math:`K_3` 
 
 	Three-variable :math:`K_3` fully-connected problem is embedded in six qubits on a D-Wave 2000Q. Variable :math:`s_0`, highlighted in dark magenta, is represented by three qubits, numbers 0, 4, and 7; Variable :math:`s_2` is represented by two qubits, numbers 3 and 6, shown with their connecting edge emphasized (and displaying a solution of :math:`+1`).
 
+.. _concepts__chain_strength:
+
 Chain Strength
 --------------
 
--- a/docs/examples/and.rst
+++ b/docs/examples/and.rst
@@ -338,7 +338,7 @@ to view the solution on the QPU.
   View of the logical and embedded problem rendered by Ocean's problem inspector. The AND gate's original QUBO is represented on the left; its embedded representation, on the right, shows a two-qubit chain of qubits 0 and 4 for variable Z. The current solution displayed, :math:`X1=1, X2=0, Z=0`, is represented by white and gold dots for binary :math:`0, 1` and white and blue dots for spin values :math:`-1, 1`.
 
 For comparison, the following code purposely weakens the 
-:ref:`chain strength <embedding_sdk>` (strength of the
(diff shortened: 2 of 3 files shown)


docs/overview/sapi.rst
```diff
@@ -92,7 +92,7 @@ runs these same configuration steps):
    file if needed.
 
 Alternatively, you can create and edit a
-:doc:`D-Wave Cloud Client configuration file </docs_cloud/sdk_index>`
+:ref:`D-Wave Cloud Client configuration file <configurationFiles>`
 manually.
```
JoelPasvolsky, line 96: Typically we prefer inter-repo links to target higher-level content because content changes over time and you get broken links. In this case, the risk is low because (1) the source is the SDK so the doc build will break if the target changes (2) the link target name is unlikely to be duplicated elsewhere and (3) that section should have a long life. 
But you'll see references such as :doc:`D-Wave Cloud Client configuration file </docs_cloud/sdk_index>` in a lot of docstrings and I'd be cautious about changing such references in other repos.

docs/concepts/embedding.rst
```diff
@@ -90,6 +90,8 @@ intentionally sets the embedding shown below to represent this same :math:`K_3`
 
 	Three-variable :math:`K_3` fully-connected problem is embedded in six qubits on a D-Wave 2000Q. Variable :math:`s_0`, highlighted in dark magenta, is represented by three qubits, numbers 0, 4, and 7; Variable :math:`s_2` is represented by two qubits, numbers 3 and 6, shown with their connecting edge emphasized (and displaying a solution of :math:`+1`).
 
+.. _chain_strength:
```
JoelPasvolsky, line 93: Because a marker like chain_strength could easily pop up elsewhere, I'd be more comfortable with one less likely to be duplicated such as `_concepts_chain_strength`

JoelPasvolsky (approved):
LGTM, thank you!

docs/overview/sapi.rst
mcfarljm: Thanks for the explanation.  For now I'd prefer to follow the usual practices.  Even though there are some safeties that you mention, it's only a minor clarity issue, so I went ahead and dropped this commit.

docs/concepts/embedding.rst
mcfarljm: Good suggestion, I went ahead and switched to `_concepts__chain_strength` in attempt to provide some indication of the doc vs section distinction.

arcondello:
This is ready to merge?

mcfarljm:
Yes, ready to merge

PR #105, opened 2020-11-12
More minor documentation updates

mcfarljm:
In the case of the NetworkX random graphs link, I think it would be the most clear to link directly to the documentation for `barabasi_albert_graph` (https://networkx.org/documentation/stable/reference/generated/networkx.generators.random_graphs.barabasi_albert_graph.html#networkx.generators.random_graphs.barabasi_albert_graph) instead of the random graphs section, but I'm not sure if that works with the github.io URL format or if there is a reason for using that instead of networkx.org.

### DIFF
--- a/docs/examples/hybrid1.rst
+++ b/docs/examples/hybrid1.rst
@@ -59,7 +59,7 @@ Formulate the Problem
 
 This example uses a synthetic problem for illustrative purposes: a NetworkX
 generated graph,
-`NetworkX barabasi_albert_graph() <https://networkx.github.io/documentation/stable/reference/generators.html#module-networkx.generators.random>`_\ , with random +1 or -1
+`NetworkX barabasi_albert_graph() <https://networkx.github.io/documentation/stable/reference/generators.html#module-networkx.generators>`_\ , with random +1 or -1
 couplings assigned to its edges.
 
 .. testcode::
--- a/docs/examples/multi_gate.rst
+++ b/docs/examples/multi_gate.rst
@@ -91,7 +91,7 @@ Formulating the Problem as a CSP
 This example demonstrates two formulations of constraints from the problem's
 logic gates:
 
-#. Single comprehensive constraint.
+1. Single comprehensive constraint.
 
 .. testcode::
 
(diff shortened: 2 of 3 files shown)


docs/examples/hybrid1.rst
```diff
@@ -59,7 +59,7 @@ Formulate the Problem
 
 This example uses a synthetic problem for illustrative purposes: a NetworkX
 generated graph,
-`NetworkX barabasi_albert_graph() <https://networkx.github.io/documentation/stable/reference/generators.html#module-networkx.generators.random>`_\ , with random +1 or -1
+`NetworkX barabasi_albert_graph() <https://networkx.github.io/documentation/stable/reference/generators.html#module-networkx.generators.random_graphs>`_\ , with random +1 or -1
 couplings assigned to its edges.
```
JoelPasvolsky, line 63: I think that suffix `.random` might have been a copy&paste error, I think I wanted `https://networkx.org/documentation/stable/reference/generators.html#module-networkx.generators`. I worry that the page structure is quite likely to change, so a level up at least is more stable. From the page users can easily search for wanted graphs

docs/examples/multi_gate.rst
```diff
@@ -93,40 +93,40 @@ logic gates:
 
 #. Single comprehensive constraint.
 
-.. testcode::
```
JoelPasvolsky, line 96: Maybe better to just switch to 
1.  Single comprehensive constraint.
...
2. Multiple small constraints.
...

There aren't going to be additional sections so no need for automated counting. Sphinx renders the examples clearly as a whole with background color so flush to the side is fine, no need to tab right.

JoelPasvolsky (commented):
Thank you!

docs/examples/hybrid1.rst
mcfarljm: Got it, I went ahead and changed to `#module-networkx.generators`

JoelPasvolsky (approved):

docs/examples/multi_gate.rst
mcfarljm: Makes sense to me.  I switched it over to manual numbering without indentation.

arcondello:
This is ready to merge?

mcfarljm:
Yes, ready to merge, thanks

PR #106, opened 2020-12-01
Build minorminer's C docs

JoelPasvolsky:
Requires minorminer release with https://github.com/dwavesystems/minorminer/commit/95fcf55233931826d47134b5bbcefd05fc01f835

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -20,6 +20,7 @@ import os
 import sys
 import subprocess
 
+config_directory = os.path.dirname(os.path.abspath(__file__))
 sys.path.insert(0, os.path.abspath('.'))
 sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
 
@@ -43,7 +44,8 @@ extensions = [
     'sphinx.ext.todo',
     'sphinx.ext.viewcode',
     'sphinx.ext.githubpages',
-    'sphinx.ext.ifconfig'
+    'sphinx.ext.ifconfig',
+    'breathe',
 ]
 
 autosummary_generate = True
@@ -129,6 +131,15 @@ import penaltymodel.lp as lp
 
 import dwave.inspector
 """
+
+
+# Path to the cpp xml files
+breathe_projects = {"minorminer": os.path.join(
+    config_directory, '../minorminer/docs/build-cpp/xml/')}
+
+breathe_default_project = "minorminer"
+
+breathe_default_members = ('members', )
 
 # -- Options for HTML output ----------------------------------------------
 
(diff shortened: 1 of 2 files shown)


docs/conf.py
```diff
@@ -277,3 +290,9 @@ def setup(app):
 
 # show inherited members
 # autodoc_default_flags = ['members', 'undoc-members', 'inherited-members', 'show-inheritance']
+
+read_the_docs_build = os.environ.get('READTHEDOCS', None) == 'True'
+
+if read_the_docs_build:
+
+    subprocess.call('cd ../minorminer/docs/; make cpp', shell=True)
```
randomir, line 298: Is there any reason not to put this in [`.readthedocs.yml`](https://docs.readthedocs.io/en/stable/config-file/v2.html)?

I would not recommend we run `make` on module import. :see_no_evil:

JoelPasvolsky:
The `ci/circleci: test-doctest` failure is due to breathe importing an obsolete sphinx extension. I set the same requirement on breathe that is in https://github.com/dwavesystems/minorminer/blob/main/docs/requirements.txt: breathe<4.13.0. This problem is fixed in breathe versions 4.14.1 and up. Locally I run doctest successfully with version 4.23.0. @boothby, what's the reason for the upper limit on breathe in minorminer? Can I up it a tad?

boothby:
Great question... git blames you for that :stuck_out_tongue_closed_eyes:

JoelPasvolsky:
Well that's embarrassing. I see that at the time we had an older version of sphinx in minorminer so probably we can update it there too. Here sphinx is not pinned at all so I'm unpinning breathe too. Thanks @boothby!

docs/conf.py
randomir: Ups. On a closer look, RtD doesn't support custom (pre-)build steps. They do have a [feature request opened](https://github.com/readthedocs/readthedocs.org/issues/6662), likely to be implemented at some point.

I expected it's possible to customize the [requirements install](https://docs.readthedocs.io/en/stable/config-file/v2.html#python) a bit more. Also, to use a custom docker image. But no. Pretty disappointing. :stuck_out_tongue: 

There is [this hackish workaround](https://github.com/readthedocs/readthedocs.org/issues/6662#issuecomment-653675848) to put a build step in a separate yaml file, but I'm not sure that's any better than running from `conf.py`.

JoelPasvolsky: Thanks @randomir!

randomir:
@JoelPasvolsky, is this ready for merge?

JoelPasvolsky:
@randomir, yup, go for it.

PR #108, opened 2020-12-10
Correction to DQM equation

### LINKED ISSUE #107: Documentation for DQM equation
The equation given for the DQM under `docs/concepts/dqm.rst` does not appear to be correct.  I believe that the linear and quadratic terms are associated with indicators on the cases of each discrete variable, not the discrete variables themselves.  As written, the equation appears to suggest we could do something like `red * green`.

Correcting this may complicate the notation quite a bit.  I suspect that four indices are needed to express the quadratic terms (two for the discrete variables, two for the cases, just like in `DiscreteQuadraticModel.set_quadratic_case`).  And if `v_i` denotes a discrete variable, it may be necessary to use a different notation in the equation, such as `x_i_j`, which would be a binary indicator for whether discrete variable `i` is set to case `j`.

mcfarljm:
DQM equation should be expressed in terms of binary indicators for the cases of each discrete variable, as opposed to the discrete variables themselves (which can have arbitrary or non-numerical values).

Here is what the new rendered equation looks like:
image

Other suggestions for improving the notation welcome.

Closes #107.

### DIFF
--- a/docs/concepts/dqm.rst
+++ b/docs/concepts/dqm.rst
@@ -5,18 +5,19 @@ Discrete Quadratic Models
 =========================
 
 The discrete quadratic model (DQM) is a polynomial over discrete variables with
-terms all of degree two or less, represented by equation,
+terms all of degree two or less.  Suppose that there are :math:`N` discrete
+variables :math:`\bf{d}_i`, each with :math:`n_i` cases.  Conceptually, the
+cases may represent any collection of discrete values, such as ``{red, green,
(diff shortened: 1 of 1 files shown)


hsadeghidw (approved):

mcfarljm:
@JoelPasvolsky do you want any further review, or is this ready to merge?

JoelPasvolsky:
LGTM, thank you!

mcfarljm:
@arcondello This should be ready to merge unless you'd like another review

PR #113, opened 2021-01-26
Fix blackduck scans

randomir:


randomir:
Tested in an SSH session, but the real test will be the night after it's merged. :ocean:

arcondello (approved):

Issue #114, opened 2021-01-27
DQM definition could be clarified


jackraymond:
**Description**
I think the definition of DQM might be clarified. It is not made clear (enough) what is the variable space, and what (if any) are the constraints. 
The current definition is here:
https://docs.ocean.dwavesys.com/en/stable/concepts/dqm.html
I make a suggestion below, but it may create other issues. On balance, I believe presenting both polynomials will allow people to make a clearer connection between their problems, and BQMs. 

Beginning from:
"... each with 𝑛𝑖 cases." A discrete quadratic model may be defined as
H(d) = \sum_{i} a(d_i) + \sum_{i,j} b_{ij}(d_i,d_j) + c (-1-)
where a() and b() are real valued functions.

We can represent any DQM by an equivalent model over binary variables, each discrete state d_i may be replaced by a vector of binary variables using one hot encoding, where exactly one binary variable is True (value 1) and all others are False (0). A discrete quadratic model is then defined by 
AS CURRENTLY WRITTEN (-2-)
subject to one hot constraints \sum_a x_{i,a} = 1 \forall i. 
Representations (1) and (2) are equivalent over the feasible space meeting the one hot encoding constraints. Representation (2) ascribes energies both to the feasible space satisfying constraints, and an infeasible space violating constraints. Representation (2) is used by ocean SDK. 

In ocean the model is defined by the latter function.  

**To Reproduce**
Stack Overflow provides an excellent guide on [how to create a Minimal, Complete and Verifiable example.](https://stackoverflow.com/help/mcve)

**Expected behavior**

**Environment:**
 - OS: [Ubuntu 16.04.4 LTS]
 - Python version: [e.g. 3.7.0]

**Additional context**

mcfarljm:
This makes sense to me, and I like the new representation (1), as it is simpler and seems more intuitive.  It might even make sense to just replace the current representation with this new one, as opposed to keeping them both.

PR #115, opened 2021-01-28
Update DQM definition 

JoelPasvolsky:
Closes https://github.com/dwavesystems/dwave-ocean-sdk/issues/114

### DIFF
--- a/docs/concepts/dqm.rst
+++ b/docs/concepts/dqm.rst
@@ -4,13 +4,31 @@
 Discrete Quadratic Models 
 =========================
 
-The discrete quadratic model (DQM) is a polynomial over discrete variables with
-terms all of degree two or less.  Suppose that there are :math:`N` discrete
-variables :math:`\bf{d}_i`, each with :math:`n_i` cases.  Conceptually, the
-cases may represent any collection of discrete values, such as ``{red, green,
-blue, yellow}`` or ``{3.2, 67}``.  Using a binary variable :math:`x_{i,u}` to
-indicate whether discrete variable :math:`\bf{d}_i` is set to case :math:`u`,
-the objective function can be expressed by the equation:
+The discrete quadratic model (DQM) is a polynomial over *discrete* variables, 
+with all terms of degree two or less, where a discrete variable represents some 
+collection of distinct values, such as ``{red, green, blue, yellow}`` or 
+``{3.2, 67}``, which are called the variable's *cases*.  
+
+A discrete quadratic model may be defined as
+
+.. math::
+
+    H(\bf{d}) = \sum_{i} a(\bf{d}_i) + \sum_{i,j} b(\bf{d}_i,\bf{d}_j) + c
+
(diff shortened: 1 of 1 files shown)


docs/concepts/dqm.rst
```diff
@@ -4,13 +4,31 @@
...
+The discrete quadratic model (DQM) is a polynomial over *discrete* variables, 
+with all terms of degree two or less, where a discrete variable represents some 
+collection of distinct values, such as ``{red, green, blue, yellow}`` or 
+``{3.2, 67}``, which are called the variable's *cases*.  
+
+A discrete quadratic model may be defined as
+
+.. math::
+
+    H(\bf{d}) = \sum_{i} a(\bf{d}_i) + \sum_{i,j} b(\bf{d}_i,\bf{d}_j) + c
```
mcfarljm, line 16: I believe that the "a" and "b" functions need subscripts:
```suggestion
    H(\bf{d}) = \sum_{i} a_i(\bf{d}_i) + \sum_{i,j} b_{i,j}(\bf{d}_i,\bf{d}_j) + c
```

docs/concepts/dqm.rst
```diff
@@ -4,13 +4,31 @@
...
+
+where :math:`\bf{d}_i` are the discrete variables, :math:`a()` and :math:`b()`
```
mcfarljm, line 18: Perhaps:
```suggestion
where :math:`\bf{d}_i` are the discrete variables, :math:`a_i()` and :math:`b_{i,j}()` 
```
?

mcfarljm (commented):
Looks good, a couple minor comments on subscripting the "a" and "b" functions

hsadeghidw:
I have some doubts about the new definition. I think that the new definition can map to a DQM.

We want to make it clear to the users that there are independent linear and quadratic biases for a combination of variables and cases.

I think one thing to add to this [definition](https://docs.ocean.dwavesys.com/en/stable/concepts/dqm.html) is that the variable picks one and only one case. Otherwise, in my opinion, the definition is clear enough.

jackraymond:
On balance I think it would be better to gradually move to a simpler definition; although I appreciate inertia may be a sufficient argument against. My thinking is here:
https://confluence.dwavesys.com/pages/viewpage.action?pageId=89505592

docs/concepts/dqm.rst
jackraymond: yes, typo

docs/concepts/dqm.rst
jackraymond: agree

jackraymond (commented):
Happy with structure implemented by JoelPavolsky, and indexing errors corrected by mcfarljm, looks improved to me.

randomir:
With indexing corrections committed, I gather, @jackraymond, you're fine with merging this?

@hsadeghidw, where do you stand on this? Not sure how to read your comments above.

randomir (approved):
This looks like an improvement to me.

randomir:
Needless to say, subsequent improvements always welcome in a follow-up PR.

PR #121, opened 2021-03-29
Link to GitHub source code

JoelPasvolsky:
Still needs a bit more work but before spending the time I'd like to get feedback on whether most people prefer such access to the GitHub code or to get the current HTML rendering. You can try it out here: https://sdk-docs.readthedocs.io/en/latest/ 

(@mcfarljm, re our discussion about the "Edit on GitHub" button not working for generated stubs.)

* Disadvantages: (1) extra complexity and will likely will need the occasional bit of maintenance (2) build time is a little longer
* Advantages: I find it very convenient to have one-click access to the source code

image

Here's what our docs currently do:
image

Here's how it work with this update:
image

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -376,7 +376,8 @@ def linkcode_resolve(domain, info):
     linespec = ""
     try:
         source, lineno = inspect.getsourcelines(obj[obj_inx])
-        linespec = "#L%d" % (lineno)
+        if obj_inx != 0:
+           linespec = "#L%d" % (lineno) 
     except Exception:
         linespec = ""
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -25,6 +25,7 @@ import os
 import sys
 import subprocess
 import inspect      
+import pkg_resources
 
 config_directory = os.path.dirname(os.path.abspath(__file__))
 sys.path.insert(0, os.path.abspath('.'))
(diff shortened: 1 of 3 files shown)


docs/conf.py
```diff
@@ -296,3 +297,79 @@ def setup(app):
...
+                               'mip': 'penaltymodel_mip'},
+              'system': 'dwave-system',
+              'embedding': 'dwave-system',
+              'tabu': 'dwave-tabu'}
+
+def linkcode_resolve(domain, info):
+    """
+    Find the URL of the GitHub source for dwave-ocean-sdk objects.
+    """
+    # Based on https://github.com/numpy/numpy/blob/main/doc/source/conf.py
```
arcondello, line 323: Probably enough code that we need to include a copy of their [LICENSE](https://github.com/numpy/numpy/blob/main/LICENSE.txt).

arcondello (commented):
This is great!

randomir (approved):
:heart:

arcondello (approved):

docs/conf.py
JoelPasvolsky: @silviacorbellaro, see https://github.com/dwavesystems/dwave-ocean-sdk/pull/121/commits/90589c1e270cc3c6f54e25107d062d6ae437796f based on our discussion.

arcondello: FWIW, I think we have historically just included it as a comment in the code.

JoelPasvolsky:
@randomir, @arcondello, ready for code review. 
The `linkcode_resolve()` function now resolves the submodule to the GitHub repo and falls back to next-level source for objects that aren't resolved so users click through to the correct repo at least.

arcondello (approved):

docs/conf.py
```diff
@@ -296,3 +302,85 @@ def setup(app):
...
+    target = fn.split("/")
+    repo = target[1] if not target[1] == "dwave" else target[2]
+
+    if repo in ["minorminer", ]:
+        fn = "https://github.com/dwavesystems/" + github_map[repo] + "/blob/main" + fn
+    elif repo == 'penaltymodel':
+        fn = "https://github.com/dwavesystems/penaltymodel/blob/master/" +  \
+             github_map[repo][info['module'].split('.')[1]] + fn  
+    else:
+        fn = "https://github.com/dwavesystems/" + github_map[repo] + "/blob/master" + fn
```
randomir, line 383: You can just [use `HEAD` to point to tip of the default branch](https://stackoverflow.com/questions/64726262/is-there-a-stable-url-to-always-get-the-default-branch-in-github).

randomir: However, shouldn't we use the _actual version_ of the file included in current Ocean, and not the latest master?

JoelPasvolsky: @randomir, I'll still clean up the updated code but looks like your proposal to use `python  setup.py egg_info` rather than `distutils.core.run_setup("setup.py")` will work.

JoelPasvolsky: Okay, now links to the version of source specified by the SDK.

docs/conf.py
```diff
@@ -296,3 +302,104 @@ def setup(app):
...
+              'minorminer': ['minorminer',],
+              'neal': ['dwave-neal',],
+              'penaltymodel': [{'cache': ['penaltymodel_cache',],
+                               'core': ['penaltymodel_core',],
+                               'lp': ['penaltymodel_lp',],
+                               'maxgap': ['penaltymodel_maxgap',],
+                               'mip': ['penaltymodel_mip',]}],
+              'system': ['dwave-system',],
+              'embedding': ['dwave-system',],
+              'tabu': ['dwave-tabu',]}
```
randomir, line 323: I'd simplify this to a `rtd-project-name` -> `gh-repo-name` map, and encode package versions separately (see below).

docs/conf.py
```diff
@@ -296,3 +302,104 @@ def setup(app):
...
+        if line.split('==')[0] == 'penaltymodel':
+            github_map['penaltymodel'][0]['core'].append(line.split('==')[1].rstrip())
+        for gh, repo in github_map.items():
+            if gh=='penaltymodel':
+               for gh, repo in repo[0].items():
+                  if repo[0].replace('_', '-') in line:
+                     github_map['penaltymodel'][0][gh].append(line.split('==')[1].rstrip())
+            else: # non penalty model
+               if repo[0] in line:
+                  github_map[gh].append(line.split('==')[1].rstrip())
```
randomir, line 338: Can be simplified a lot by using `pkg_resources` to fetch project's requirements:
```python
import pkg_resources
reqs = pkg_resources.get_distribution('dwave-ocean-sdk').requires(extras=['all'])
pkgs = [pkg_resources.get_distribution(req) for req in reqs]
versions = {pkg.project_name: pkg.version for pkg in pkgs}
```
This produces:
```
>>> versions
{'dimod': '0.9.13',
 'dwave-cloud-client': '0.8.4',
 'dwave-greedy': '0.1.2',
 'dwave-hybrid': '0.6.1',
 'dwave-inspector': '0.2.5',
 ...}
```

Note that we have to use `all` extras to include both `penaltymodel-maxgap` and `penaltymodel-mip`. Also, for this to work, versions of packages installed in the current venv have to match those required by `install_requires`. Currently, RtD is set-up to install docs deps with `pip install -r docs/requirements.txt` and package deps via `python setup.py install`. This might result in a version mismatch, and we should install all deps using the same method, e.g. pip. To do this, I'd recommend (and we should do it regardless) to configure RtD build through `.readthedocs.yml`, where we would have a section like this:
```
python:
   version: 3.7
   install:
      - method: pip
        path: .
      - requirements: docs/requirements.txt
```

docs/conf.py
JoelPasvolsky: Done

docs/conf.py
JoelPasvolsky: Replaced `python setup.py egg_info` with `pkg_resources`. Still need to update `.readthedocs.yml`

randomir (approved):
:tada:

docs/conf.py
JoelPasvolsky: The recommendation to use a YAML file is implemented with https://github.com/dwavesystems/dwave-ocean-sdk/pull/123

PR #122, opened 2021-04-01
Remove support for Python 3.5 from installation instructions

JoelPasvolsky:


arcondello (approved):

randomir (approved):
Hah! If only removing py35 support was as easy as a 1 char commit. :rofl:

PR #123, opened 2021-04-12
Add YAML file for RTD

JoelPasvolsky:


.readthedocs.yml
```diff
@@ -0,0 +1,16 @@
...
+sphinx:
+    builder: html
+    configuration: docs/conf.py
+
+python:
+   version: 3.7
+   install:
+      - requirements: docs/requirements.txt
+      - method: pip
+        path: .
```
randomir, line 15: I'd invert those two, as docs deps naturally extend package deps.

randomir (approved):
LGTM!

.readthedocs.yml
JoelPasvolsky: Done, thanks!

.readthedocs.yml
```diff
@@ -0,0 +1,16 @@
...
+sphinx:
+    builder: html
+    configuration: docs/conf.py
+
+python:
+   version: 3.7
+   install:
+      - method: pip
+      - requirements: docs/requirements.txt
+        path: .
```
randomir, line 15: Syntax error. RtD build should fail on this. `path` has to go together with `method`.

arcondello (approved):

PR #131, opened 2021-06-14
For consideration: replace frontpage table with panels

JoelPasvolsky:
What do you think of this format?

image

You can try it here: https://sdk-docs.readthedocs.io/en/latest/ (don't know why but bullets are broken in that build)


hhtong:
Personally I like being able to see the summaries of the packages immediately and it seems like an additional two steps to have to go into the ‘Documentation & Descriptions’ tab and then open the collapsed summaries individually. I’m wondering if maybe it’d be better to have them open by default, though the page gets longer of course. I like the look of the links a lot more though!

JoelPasvolsky:
I found the issue breaking bullets (unrelated to this PR, so needed for next doc build of the SDK regardless) and will make a PR to fix that tomorrow.

arcondello:
I agree with Heidi, I think while we have separate packages it makes sense to lead with the descriptions.

JoelPasvolsky:
My considerations were that (1) the descriptions are useful for newbies and will be used maybe a couple of times (so two clicks is not a huge deal) while the one-click links to source code are useful for everyone long-term (2) brings links to system documents up into view.
But if these considerations are not strong enough I can switch the tabs and set the default to open (basically having them open by default makes the panels a formatting, no one will close them)

JoelPasvolsky:
If we want to open on Documentation tab with descriptions visible, then a better formatting would be this below. To keep the panels a decent size, we'd need to stick to one-sentence descriptions here.
image

JoelPasvolsky:
Here's how @randomir's proposal would look using sphinx-panels (more or less):
image

randomir:
Can the docs/code links be clicked (or does that expand the panel)?

JoelPasvolsky:
@randomir, if your mouse is over a badge, clicking opens its link target. Clicking anywhere else on the panel, expands the panel

hhtong:
👍 I'm happy with how that looks and I prefer having the docs/code links next to each other like this as well

JoelPasvolsky:
You can test run the latest here: https://sdk-docs.readthedocs.io/en/latest/ 
@hbarovertwo, @hhtong, @randomir, @arcondello   
I still need to add the MIT license

docs/index.rst
```diff
@@ -32,72 +32,104 @@ problems with quantum computers.
...
+.. dropdown:: **dwave-inspector** :link-badge:`https://docs.ocean.dwavesys.com/en/stable/docs_inspector/sdk_index.html, "docs", cls=badge-primary text-white` :link-badge:`https://github.com/dwavesystems/dwave-inspector, "code", cls=badge-dark text-white` Visualizer for problems submitted to quantum computers.
+
+   A tool for visualizing problems submitted to, and answers received from, a D-Wave
+   structured solver such as a D-Wave 2000Q quantum computer.
+
+.. dropdown:: **dwave-neal** :link-badge:`https://docs.ocean.dwavesys.com/en/stable/docs_neal/sdk_index.html, "docs", cls=badge-primary text-white` :link-badge:`https://github.com/dwavesystems/dwave-neal, "code", cls=badge-dark text-white` Simulated annealing sampler.
+
+   An implementation of a simulated annealing sampler.
+
+.. dropdown:: **dwave-networkx** :link-badge:`https://docs.ocean.dwavesys.com/en/stable/docs_dnx/sdk_index.html, "code", cls=badge-primary text-white` :link-badge:`https://github.com/dwavesystems/dwave-networkx, "code", cls=badge-dark text-white` NetworkX extension.
```
randomir, line 73: two `code` pills

docs/index.rst
```diff
@@ -32,72 +32,104 @@ problems with quantum computers.
...
+
+   dwave-networkx provides tools for working with :term:`Chimera` graphs and implementations of
+   graph-theory algorithms on the D-Wave system and other binary quadratic model
+   :term:`sampler`\ s.
+
+.. dropdown:: **dwave-ocean-sdk** :link-badge:`https://docs.ocean.dwavesys.com/en/stable/index.html, "docs", cls=badge-primary text-white` :link-badge:`https://github.com/dwavesystems/dwave-ocean-sdk, "code", cls=badge-dark text-white` Ocean software development kit.
+
+   Installer for D-Wave's Ocean Tools.
+
+.. dropdown:: **dwave-preprocessing** :link-badge:`https://docs.ocean.dwavesys.com/en/stable/docs_preprocessing/sdk_index.html, "code", cls=badge-primary text-white` :link-badge:`https://github.com/dwavesystems/dwave-preprocessing, "code", cls=badge-dark text-white` Preprocessing tools for quadratic models.
```
randomir, line 86: another two `code` pills

randomir (commented):
This now looks great. I have just two minor comments:
- pills/badges should use a different on-hover text color. Currently it's black, and when the dark "code" pill is hovered, the text is illegible.
- pills are jagged, a little. That is, they form this jagged vertical line. Could they be moved/aligned to the far right or far left? e.g.
  ```
  <package> - <tagline> (docs) (code)
  ```

JoelPasvolsky:
@randomir, I can swap the short description and the badges and I could also change the code badge's color from black to grey for slightly better when hovering. Visibility of the text when hovering seems a minor concern given that (1) one reads the text before pointing one's mouse on the badge and (2) one sees the same text in the surrounding dark badges. This moves the "jagged" (I see a nice if a bit noisy sine wave) badge line to the end of the text:

image
 
Is this better?

randomir:
Yes, I think this is better. It would be perfect if buttons could be aligned (floated) right. But it's fine if this is too hard to do.

JoelPasvolsky:
Thanks, @randomir, I've moved the text/badges and updated the color. I expect aligning right to be possible but will likely add quite a bit of complexity because it's not a property in RST, so I'll see tomorrow if there's a reasonable method (I'd prefer not to add a bunch of CSS files that we'd need to maintain for this).

randomir:
Thanks, @JoelPasvolsky.

If there's no simple way to achieve this:
image

adding these two lines of CSS to the `badge` class shouldn't be too bad:
```
float: right;
margin-left: 0.25rem;
```
(ideally, we can create a new class (`floated-badge`) with these two rules and add it to our badges)

arcondello (approved):
Looks great! Subject to whatever tweaks you and Rad want to work out of course.

docs/_static/theme_overrides.css
```diff
@@ -9,3 +9,17 @@
...
+  float: right;
+  margin-left: 0.25rem;
+}
+
+.badge-primary.fr
+{
+  float: right;
+  margin-left: 0.25rem;
+}
+
```
randomir, line 25:
```suggestion
.badge-primary.fr, .badge-secondary.fr {
  float: right;
  margin-left: 0.25rem;
}
```

randomir: A separate question, why do we need the `fr` class?

docs/index.rst
```diff
@@ -32,72 +32,136 @@ problems with quantum computers.
...
+   An implementation of a simulated annealing sampler.
+
+.. dropdown:: **dwave-networkx** --- NetworkX extension. \
+   :link-badge:`https://docs.ocean.dwavesys.com/en/stable/docs_dnx/sdk_index.html, "docs", cls=badge-primary fr text-white` \
+   :link-badge:`https://github.com/dwavesystems/dwave-networkx, "code", cls=badge-secondary fr text-white` 
+
+   Extension of NetworkX—a Python language package for exploration and analysis
+   of networks and network algorithms—for users of D-Wave Systems.
+
+   dwave-networkx provides tools for working with :term:`Chimera` graphs and implementations of
```
hhtong, line 94: and :term:`Pegasus` graphs?

hhtong (approved):
LGTM!

silviacorbellaro (approved):

docs/index.rst
JoelPasvolsky: Good catch @hhtong, added.

randomir (approved):
LGTM.

One minor detail you might care about is the order of the badges -- which now reversed when rendered vs. written, due to the nature of `float: right` rule.

JoelPasvolsky:
I did think about that and it seems to me nicer to have the blue at the very right rather than dark-blue-dark.

Issue #132, opened 2021-06-15
Sphinx silently undoes bullet formatting


JoelPasvolsky:
**Description**
Sphinx with the latest docutil ignores some formats such as bullets:

![image](https://user-images.githubusercontent.com/34041130/122069327-f647cd80-cda9-11eb-8419-344b51343903.png)

**To Reproduce**
Build ocean docs with latest sphinx

**Expected behavior**
Keep specified formatting

**Environment:**
RTD

**Additional context**
Needed before next SDK release

PR #133, opened 2021-06-15
Pin docutil to prevent unformatting bullets etc

JoelPasvolsky:
Fixes https://github.com/dwavesystems/dwave-ocean-sdk/issues/132


randomir (approved):
LGTM

PR #139, opened 2021-06-29
Update ping to select qpu in example

JoelPasvolsky:


docs/docs_cli.rst
```diff
@@ -142,7 +142,7 @@ configured using the :code:`dwave setup` or :code:`dwave config` commands.
 
 .. code-block:: bash
 
-    $ dwave ping
+    $ dwave ping -s {"qpu": true}
```
randomir, line 145:
```suggestion
    $ dwave ping -s '{"qpu": true}'
```
Also, this will only work in Bash, not cmd/powershell.

randomir (changes requested):

randomir (approved):
LGTM!

PR #142, opened 2021-08-26
Use dimod gates in examples (1/2)

JoelPasvolsky:
Requires dimod 0.10.x. This PR replaces uses of `dwavebinarycsp.factories.constraint.gates` when used only to provide a Boolean gate, not when dwavebinarycsp is used for additional constraint creation.


JoelPasvolsky:
Rebased to sdk 3.5.0 (@randomir)

hhtong (approved):

PR #145, opened 2021-09-03
Restructure programming model files

JoelPasvolsky:
Interacts with https://github.com/dwavesystems/dwave-ocean-sdk/pull/142


JoelPasvolsky:
Rebased to sdk 3.5.0 (@randomir)

docs/concepts/index.rst
```diff
@@ -214,10 +223,14 @@ Glossary
...
 
       Penalty model
-         An approach to solving constraint satisfaction problems (CSP) using an :term:`Ising` model 
-         or a :term:`QUBO` by mapping each individual constraint in the CSP to a ‘small’ Ising model 
+         An approach to solving constraint satisfaction problems (CSP) using an :term:`Ising` model
+         or a :term:`QUBO` by mapping each individual constraint in the CSP to a ‘small’ Ising model
          or QUBO.
 
+      Quadratic model
+         A collection of binary and/or integer variables with associated linear
```
arcondello, line 231: I think it's actually better to not restrict the variable types. Quadratic model just means that you have a set of variables with linear and quadratic relationships. Conceivably it could include continuous variables in the future, for instance.

docs/concepts/penalty.rst
```diff
@@ -0,0 +1,78 @@
+.. _penalty_sdk:
+
+==============
+Penalty Models
+==============
+
+Penalty models represent constraints as small models such as :ref:`bqm_sdk`\ s
+that have higher values for infeasible states (values of variables that violate
```
arcondello, line 8: `values`? Seems like a slightly strange word in this context. I suppose you're trying to avoid "energies"?

docs/overview/formulation.rst
```diff
@@ -0,0 +1,163 @@
...
+============================================
+Formulation: Objectives and Quadratic Models
+============================================
+
+For quantum computing, as for classical, solving a problem requires that it
+be formulated in a way the computer and its software understand.
+
+For example, if you want your laptop to calculate the area of a $1 coin, you might
+express the problem as an equation, :math:`A=\pi r^2`, that you program as
+:code:`math.pi*13.245**2` in your Python CLI. For a laptop with Python software,
```
arcondello, line 12: "CLI" seems like an abbreviation at a totally different altitude than the rest of this section.

docs/concepts/index.rst
JoelPasvolsky: Changed to "A collection of variables with associated linear and "

docs/overview/formulation.rst
```diff
@@ -0,0 +1,163 @@
...
+
+.. _formulating_cqm:
+
+Example: CQM for Greatest Rectangle Area
+========================================
+
+Consider the simple problem of finding the rectangle with the greatest area when the
+circumference is limited.
+
+In this example, the circumference of the rectangle is set to 8 (meaning the
```
arcondello, line 101: I think "circumference" usually is only applied to a circle, "perimeter"?

arcondello (approved):
LGTM, aside from a few minor nits

docs/concepts/penalty.rst
JoelPasvolsky: Yes, for the more abstract models---outside the context of a particular physical system such as a BQM embedded on a QPU where energies are straightforward---users might be more familiar with a generic cost function interpretation.

arcondello: I would probably use the word "cost" then. "value" seems overloaded to me but :shrug:

docs/overview/formulation.rst
JoelPasvolsky: --> terminal

PR #156, opened 2021-10-04
Fix some typos in class names

JoelPasvolsky:


randomir (approved):
lgtm

JoelPasvolsky:
doctest failure because using dwave-system wheel version 1.10.0 that was prior to https://github.com/dwavesystems/dwave-system/pull/429

PR #167, opened 2021-11-05
Update SAPI access

JoelPasvolsky:
You can see the rendered HTML here: https://sdk-docs.readthedocs.io/en/latest/overview/sapi.html

### DIFF
--- a/docs/docs_cli.rst
+++ b/docs/docs_cli.rst
@@ -76,13 +76,13 @@ for a full setup.
 
     Optionally install non-open-source packages and configure your environment.
 
-    Do you want to select non-open-source packages to install (y/n)? [y]:
+    Do you want to select non-open-source packages to install (y/n)? [y]: ↵
 
     D-Wave Drivers
     These drivers enable some automated performance-tuning features.
     This package is available under the 'D-Wave EULA' license.
     The terms of the license are available online: https://docs.ocean.dwavesys.com/eula
-    Install (y/n)? [y]:
+    Install (y/n)? [y]: ↵
     Installing: D-Wave Drivers
     Successfully installed D-Wave Drivers.
 
@@ -90,7 +90,7 @@ for a full setup.
     This tool visualizes problems submitted to the quantum computer and the results returned.
     This package is available under the 'D-Wave EULA' license.
     The terms of the license are available online: https://docs.ocean.dwavesys.com/eula
-    Install (y/n)? [y]:
+    Install (y/n)? [y]: ↵
     Installing: D-Wave Problem Inspector
     Successfully installed D-Wave Problem Inspector.
 
@@ -99,9 +99,9 @@ for a full setup.
     Try 'dwave config create --full' for more options.
 
     Creating new configuration file: /home/jane/.config/dwave/dwave.conf
-    Profile [defaults]:
-    Updating existing profile: defaults
-    Authentication token [skip]: ABC-1234567890abcdef1234567890abcdef
+    Profile [defaults]: ↵
+    Updating existing profile: defaults ↵
+    Authentication token [skip]: ABC-1234567890abcdef1234567890abcdef ↵
     Configuration saved.
 
 .. cli-example-setup-end-marker
@@ -124,9 +124,9 @@ The output shown below includes the interactive prompts and placeholder replies.
     Try 'dwave config create --full' for more options.
 
     Creating new configuration file: /home/jane/.config/dwave/dwave.conf
-    Profile [defaults]:
-    Updating existing profile: defaults
-    Authentication token [skip]: ABC-1234567890abcdef1234567890abcdef
+    Profile [defaults]: ↵
+    Updating existing profile: defaults ↵
+    Authentication token [skip]: ABC-1234567890abcdef1234567890abcdef ↵
     Configuration saved.
 
 .. cli-example-config-end-marker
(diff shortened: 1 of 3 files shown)


docs/overview/sapi.rst
```diff
@@ -89,22 +258,29 @@ runs these same configuration steps):
...
+You can always set or override your cofigurations in your code or with environment
+variables. For more information, see the examples in this document or
+:doc:`D-Wave Cloud Client </docs_cloud/sdk_index>`.
+
+.. _sapi_access_config_verify:
 
-Verifying Your Configuration
-----------------------------
+Verify Your Configuration
+=========================
```
randomir, line 283: Instead of 
```
dwave ping -s '{"qpu": true}'
```
maybe it's nicer to write:
```
dwave ping --client qpu
```

It's not ideal, and we might add `--qpu` switch (et al).

randomir: Unrelated, I think using actual endpoint/solver in the example output would be much clearer than using fake ones.

docs/overview/sapi.rst
```diff
@@ -131,80 +307,66 @@ You can test that your solver access is configured correctly with the
...
+You can set up your configuration file with profiles for multiple regions, as
+shown below.
+
+.. code-block:: bash
+
+  [defaults]
+  token = ABC-123456789123456789123456789
+
+  [europe]
+  token = DEF-987654321987654321987654321
```
randomir, line 366: It might be confusing/misleading to re-set the token. Tokens are global, not per region.

randomir (commented):
Looking pretty good.

docs/overview/sapi.rst
```diff
@@ -1,62 +1,231 @@
...
+          - Not recommended outside of testing (for security reasons).
+
+     .. note:: For non-Ocean clients, you set your token in the HTTP header
+        :code:`X-Auth-Token`; see the :std:doc:`sysdocs_gettingstarted:doc_rest_api`
+        guide for information.
+
+  .. [#]
+    When you work in D-Wave's `Leap <https://cloud.dwavesys.com/leap/>`_ IDE,
+    SAPI information such as your API token is pre-configured in the default
+    workspace's environment variables.
```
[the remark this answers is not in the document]
JoelPasvolsky: Thanks. For this general content I'll keep it functional, and the technical details can go into the LeapIDE docs that are now being looked at. I've updated to:

D-Wave's `Leap <https://cloud.dwavesys.com/leap/>`_ integrated development 
environment (IDE) automatically reads your API token from your account. 
You can use the configuration options of the :ref:`table_token_config` table
to substitute an alternative token, for example, if you belong to multiple 
projects.

docs/overview/sapi.rst
```diff
@@ -1,62 +1,231 @@
...
+          - For users that belong to multiple projects, solvers accessible to the
+            current project are displayed; switch the current project if needed.
+          - Solvers may be local to a region.
+        * - :ref:`dwave-cloud-client <sdk_index_cloud>`
+          - Use the :ref:`dwave CLI <dwave_cli>` tool or Ocean's
+            :meth:`~dwave.cloud.client.Client.get_solvers` method.
+          - For users that belong to multiple projects, queries return solvers
+            accessible to the current API token; switch to another project's API
+            token if needed.
+          - Solvers may be local to a region; see the :ref:`sapi_intro_multiregion`
```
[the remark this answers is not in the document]
JoelPasvolsky: I was thinking of hybrid or SW solvers, which possibly could have replicas across the regions or be imported, no? That is, I used the may so as not to preclude the option multiple regions might host such a solver with the same name or enable remote access. I don't know if we'll have such a use cases or not.

docs/overview/sapi.rst
```diff
@@ -1,62 +1,231 @@
...
+            section to query solvers outside your default region.
+
+     Example: using :ref:`dwave CLI <dwave_cli>` to see the available solvers,
+     their parameters, and properties (the output shown below is illustrative only).
+
+     .. include:: ../docs_cli.rst
+        :start-after: cli-example-solvers-start-marker
+        :end-before: cli-example-solvers-end-marker
```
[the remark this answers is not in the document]
JoelPasvolsky: I agree. I was tempted to replace this example imported from cloud-client with a `dwave solvers --al` example, and I do almost that in the "Example: Viewing a Region's Solvers" section. I suggest we update the example in cloud client and then it will be brought in here too

docs/overview/sapi.rst
```diff
@@ -89,22 +258,29 @@ runs these same configuration steps):
...
+   prompts except the API token\ [#]_\ . You can in the future update the file
+   if needed.
+
+.. [#] Users of on-premises systems should also enter the SAPI endpoint. Users
+   interested in using solvers hosted outside their default region can also
+   configure the SAPI endpoint (see the :ref:`sapi_intro_multiregion` section for
+   the recommended way of configuring such access).
 
 Alternatively, you can create and edit a
 :doc:`D-Wave Cloud Client configuration file </docs_cloud/sdk_index>`
```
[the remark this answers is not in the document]
JoelPasvolsky: For both of these I'm linking to cloud-client generically because many of this content's users will be first-time installers so probably best not to send them deep into cloud-client. That is why most the content uses dwave CLI and this alternative is just mentioned for completeness.

They see the `dwave config ls` a few lines above:
image

docs/overview/sapi.rst
JoelPasvolsky: Updated

docs/docs_cli.rst
```diff
@@ -99,9 +99,9 @@ for a full setup.
     Try 'dwave config create --full' for more options.
 
     Creating new configuration file: /home/jane/.config/dwave/dwave.conf
-    Profile [defaults]:
-    Updating existing profile: defaults
-    Authentication token [skip]: ABC-1234567890abcdef1234567890abcdef
+    Profile [defaults]: ↵
+    Updating existing profile: defaults ↵
```
randomir, line 103: This is not a prompt, just info output.

JoelPasvolsky: some overenthusiasm on my part. Removed in both places

randomir (approved):
LGTM!

PR #174, opened 2021-12-02
Add a banner to the README

arcondello:
I also added badges for docs and python versions, because why not.

Preview in light mode
----------------------------
image

Preview in dark mode
----------------------------
image


hhtong (approved):
That was fast!

randomir (commented):
We need a nicer banner.

Ideally, a new Ocean logo to start with, but at the very least we need to fix the SDK suffix. It simply does not fit next to Ocean text.  I would make it (1) lighter, (2) thinner, (3) increase font size _or_ use lowercase "sdk".

The meaningless curves at the bottom actually are apt here, as one can imagine they represent an ocean, I guess.

arcondello:
From Aaron:
image

randomir:
@JoelPasvolsky, that's a great point. Do we? I don't think we do.

arcondello:
Version with only ocean:

image

image

JoelPasvolsky (approved):

randomir (approved):

PR #176, opened 2021-12-02
README update

JoelPasvolsky:
Will rebase on https://github.com/dwavesystems/dwave-ocean-sdk/pull/174
Just a first draft, not ready for review, but opening to suggestions


README.rst
```diff
@@ -34,8 +35,72 @@ To install from source:
...
+
+Example Quantum Program
+-----------------------
+
+The following lines of code solve and visualize a
+`random <https://docs.ocean.dwavesys.com/en/stable/docs_dimod/reference/generators.html>`_
+problem on a quantum computer.
+
+>>> from dwave.system import DWaveSampler, EmbeddingComposite
+>>> from dimod.generators import ran_r
```
arcondello, line 67: I wonder if there is a more explicitly named problem generator we can use. Maybe `maximum_independent_set()`? Or even `gnm_random_bqm()`? The latter is at least following NumPy's naming convention.

I find `ran_r` a bit inscrutable.

README.rst
```diff
@@ -34,8 +35,72 @@ To install from source:
...
+Support
+=======
+
+Find support here:
+
+* `Leap user community <https://support.dwavesys.com/hc/en-us/community/topics>`_
+  to converse with a large community of D-Wave users.
+* `Leap help center <https://support.dwavesys.com/hc/en-us>`_
+  to search the Leap knowledge base.
+* `GitHub issues <https://github.com/dwavesystems>`_ to open issues or request
```
arcondello, line 92: Maybe link to the sdk itself or https://github.com/dwavesystems/dwave-ocean-sdk/issues

README.rst
```diff
@@ -34,8 +35,72 @@ To install from source:
...
+>>> sampleset = sampler.sample(bqm, num_reads=100)
+>>> dwave.inspector.show(sampleset)
+
+The left side of the visualized solution represents the problem's variables as
+circles, with white dots for variables assigned values of -1 and blue dots for
+values of +1; the colors of the connecting lines represent values of the
+quadratic coefficients for each pair of variables. The right side shows the qubits
+representing these variables on a quantum processing unit.
+
+.. image:: docs/_static/inspector_bqm_ran_r_30.png
```
arcondello, line 81: I love using the inspector in the README, but IMO a smaller problem where the LHS of the image is a bit less busy would look better.

README.rst
```diff
@@ -34,8 +35,72 @@ To install from source:
...
+Example Quantum Program
+-----------------------
+
+The following lines of code solve and visualize a
+`random <https://docs.ocean.dwavesys.com/en/stable/docs_dimod/reference/generators.html>`_
+problem on a quantum computer.
+
+>>> from dwave.system import DWaveSampler, EmbeddingComposite
+>>> from dimod.generators import ran_r
+>>> import dwave.inspector
```
arcondello, line 68: ok, I'm going to be ultra nitpicky here, I think the order should be
```python
>>> import dwave.inspector
>>> from dimod.generators import ran_r
>>> from dwave.system import DWaveSampler, EmbeddingComposite
```
`from` after `import` and then alphabetical.

randomir:
Already looks so much better then the original. :tada:

README.rst
JoelPasvolsky: I was thinking that the higher-level link would let users open an issue in the relevant package repo, at least for now.

README.rst
```diff
@@ -34,8 +35,72 @@ To install from source:
...
+...
+>>> bqm = ran_r(1, 30)
+>>> sampler = EmbeddingComposite(DWaveSampler())
+>>> sampleset = sampler.sample(bqm, num_reads=100)
+>>> dwave.inspector.show(sampleset)
```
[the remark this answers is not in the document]
JoelPasvolsky: Sure, I added `. code-block:: python` and updated the code

README.rst
JoelPasvolsky: This is the best replacement I've found so far, `gnp_random_bqm(25, .25, 'BINARY')`, and I find the current graphic better looking. The networkx-based problems need a bit more setup, which I'd prefer to avoid. Note the link is to the dimod generators page, so if curious users look up `ran_r`, there's a benefit of acquaintance with our generators.  

image

arcondello: Fair enough, thanks for trying

README.rst
```diff
@@ -17,11 +17,14 @@
...
+`D-Wave's <https://www.dwavesys.com>`_ suite of tools for solving hard problems
+with quantum computers.
 
+.. index-end-marker
 
 Installation
-------------
+============
 
 .. installation-start-marker
```
arcondello, line 30: I would drop the "3.6+" from this line, avoid the repeated maintenance and we have the badge above if the user wants a fast way to check.

arcondello: I would also just drop the "install from source" part altogether. Partly because I believe `pip install .` is considered the preferred method these days, and partly because I don't think the target audience of the README needs this info. I think the contributors guide might be a better place for that kind of information.

JoelPasvolsky: > drop the "install from source" part altogether
Are you thinking to apply this policy across all packages or in particular to the SDK?

arcondello: I would say everywhere, but the SDK especially since there isn't really a local package to install

arcondello (approved):
Looks *so* much better!

README.rst
randomir: I always preferred pip install over setup.py, but have been including setup.py line for consistency.

README.rst
randomir: If I was a kind of user that needs guidance on how to file Ocean issues, I would be lost with this link.

IMO, we should either link as Alex suggests, or link to `dwavesystems`, but change wording.

README.rst
```diff
@@ -17,30 +17,97 @@
...
+
+* `D-Wave website <https://www.dwavesys.com/learn/resource-library>`_ for
+  whitepapers and additional resources.
+
+Example Quantum Program
+-----------------------
+
+The following lines of code solve and visualize a
+`random <https://docs.ocean.dwavesys.com/en/stable/docs_dimod/reference/generators.html>`_
+problem on a quantum computer.
```
randomir, line 69: I understand it's easier to use a random problem for this short example, but can we also add a sentence that points to some real-world / application problems. Perhaps your getting started examples in the docs?

README.rst
```diff
@@ -17,30 +17,97 @@
...
+  bqm = dimod.generators.ran_r(1, 20)
+  sampler = dwave.system.EmbeddingComposite(dwave.system.DWaveSampler())
+  sampleset = sampler.sample(bqm, num_reads=100)
+  dwave.inspector.show(sampleset)
+
+The left side of the visualized solution represents the problem's variables as
+circles, with white dots for variables assigned values of -1 and blue dots for
+values of +1; the colors of the connecting lines represent values of the
+quadratic coefficients for each pair of variables. The right side shows the qubits
+representing these variables on a quantum processing unit.
```
randomir, line 86: I feel this whole explanation will be lost on the uninitiated reader. Perhaps we shouldn't even bother explaining these details here, but instead point interested users to the [inspector intro docs](https://docs.ocean.dwavesys.com/projects/inspector/en/latest/intro.html), or similar?

randomir: OTOH, without this paragraph, the example section would be pretty bare bones (code + image), so probably better to just leave it in.

randomir (approved):

README.rst
JoelPasvolsky: I turned the `visualized` into a link to https://docs.ocean.dwavesys.com/en/stable/docs_inspector/intro.html so the uninitiated interested in initiation can easily access more info

README.rst
JoelPasvolsky: Added
image

README.rst
JoelPasvolsky: Changed to 
```
`SDK GitHub repo <https://github.com/dwavesystems/dwave-ocean-sdk/issues>`_ to
open issues or request features on the Ocean SDK or on any one of its
`packages <https://github.com/dwavesystems>`_.
```

arcondello:
I think this is ready to merge then?

JoelPasvolsky:
Given that everyone ignored my "Just a first draft, not ready for review" and reviewed, I'll remove the "WIP", we can merge, and can later add any new content if we think of any in another PR.

README.rst
randomir: How about we also link "problem" in that sentence with https://docs.ocean.dwavesys.com/en/stable/concepts/bqm.html#bqm-sdk?

randomir: I would love if we could link from code block -- so `DWaveSampler`, `EmbeddingComposite`, etc..

README.rst
randomir: Feels a bit repetitive of the bullet list above?

randomir:
Just to be clear, I'm fine with merging this. This is great improvement.

README.rst
JoelPasvolsky: I linked problem to `https://docs.ocean.dwavesys.com/en/stable/concepts/bqm.html`
My guess is that users will look at this to get an overall impression of what problem-submission code looks like, this is not likely to be a launching point for getting more deeply acquainted with Ocean code entities.

README.rst
JoelPasvolsky: Yes, but we also wanted to follow the random example with "real-world / application problems ... your getting started examples". I think users are less annoyed by a single repetitive sentence at the end of a chunk of text than frustrated by not finding breadcrumbs to the next place they are likely to want to visit

randomir: Sounds good.

PR #177, opened 2021-12-03
Use setup.cfg rather than setup.py

arcondello:
We can almost get rid of `setup.py` altogether, but we have been telling people to use `python setup.py install` for a while so we're a bit stuck with it.


dwaveoceansdk/package_info.py
```diff
@@ -12,7 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = '4.2.0'
-__author__ = 'D-Wave Systems Inc.'
+# This file is kept for backwards compatibility
```
JoelPasvolsky, line 15: Is it worth noting deprecation in this file so users don't add code that relies on it?

arcondello: IMO there is no harm keeping it for now, but totally fine to add a deprecation warning if we think we should.

Part of my reasoning for not bothering is I don't think we're ultimately going to keep the `dwaveoceansdk` namespace at all.

setup.cfg
```diff
@@ -1,2 +1,53 @@
...
+    dwave-networkx==0.8.10
+    dwave-preprocessing==0.3.1.post0
+    dwave-qbsolv==0.3.3
+    dwave-system==1.10.0
+    dwave-tabu==0.4.2
+    dwavebinarycsp==0.1.3
+    minorminer==0.2.6
+    penaltymodel-cache==0.4.4
+    penaltymodel-maxgap==0.5.5; platform_machine != "x86_64" and platform_machine != "amd64" and platform_machine != "AMD64"
+    penaltymodel-mip==0.2.5; platform_machine == "x86_64" or platform_machine == "amd64" or platform_machine == "AMD64"
```
randomir, line 43: Do you happen to know which version of pip/setuptools this requires?

setup.py
```diff
@@ -12,74 +12,6 @@
...
-    url='https://github.com/dwavesystems/dwave-ocean-sdk',
-    long_description=open('README.rst').read(),
-    classifiers=classifiers,
-    python_requires=python_requires,
-    license='Apache 2.0',
-    packages=packages,
-    install_requires=install_requires,
-    extras_require=extras_require
-)
+setup()
```
randomir, line 17: Is the GitHub dependency tracking/scanning bug fixed?

setup.cfg
arcondello: I don't but I'll add https://github.com/dwavesystems/dimod/blob/24e06a82b0caf7128a5bc92ae758863359947bf6/pyproject.toml#L3 since that's required by our dependencies anyway.

setup.py
arcondello: Haha, I was following you https://github.com/randomir/dwave-greedy/commit/6b5657431a36cbeea7e068c2e470741efe62c1af

But actually poking around e.g. dwave-tabu https://github.com/dwavesystems/dwave-tabu/network/dependencies it looks like maybe not yet.

I am fine to leave them in the setup.py but it does feel like github has or will fix this eventually and while I like the dependency feature I am not sure it actually gets used by anything?

arcondello: Last comment on https://github.community/t/dependency-graph-does-not-support-setup-cfg-for-python/2576/7 which points to https://github.com/dependabot/dependabot-core/pull/2281#discussion_r516652962 seems relevant.

setup.cfg
arcondello: Looks like [36.2.0](https://setuptools.pypa.io/en/latest/history.html#v36-2-0), Jul 2017.

setup.py
randomir: So it's fixed upstream, but maybe not yet in GH?

We use Ocean dependents, but this should not affect that list. Probably fine to drop it.

setup.cfg
randomir: Thanks for checking.

arcondello: Just kidding, [36.2.7](https://setuptools.pypa.io/en/latest/history.html#v36-2-7), August 2017

arcondello:
I am going to go ahead and merge so I can update the dimod version. Any subsequent comments I can fix as a followup PR.

PR #178, opened 2021-12-04
Update PyQUBO version to 1.1.0

kotarotanahashi:
PyQUBO is updated so that it contains wheels for Apple Silicon.
https://github.com/recruit-communications/pyqubo/releases/tag/1.1.0


arcondello:
Hi @kotarotanahashi , we just merged a refactor of our setup.py (#177) that conflicts with this PR, can you rebase?

Also, we would like the sdk to also support aarch64 linux and python 3.10. See https://github.com/dwavesystems/dwave-preprocessing/pull/38 for an example of how we implemented it. Though I believe you use github workflows rather than circleci.

kotarotanahashi:
@arcondello 
I will create a new pull request for the current master branch.

> Also, we would like the sdk to also support aarch64 linux and python 3.10. See dwavesystems/dwave-preprocessing#38 for an example of how we implemented it. Though I believe you use github workflows rather than circleci.

OK, I will add wheel for aarch64 and Python 3.10 in the next version.

PR #184, opened 2022-01-17
Update dimod to 0.10.11

arcondello:
Should fix the dimod doctest failures.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -25,7 +25,7 @@ description = Software development kit for open source D-Wave tools
 
 [options]
 install_requires =
-    dimod==0.10.10
+    dimod==0.10.11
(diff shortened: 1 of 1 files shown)


randomir (approved):
Doctest fix looks good, but we should probably first drop py36 (in a separate PR?)

arcondello:
since dimod 0.10.11 doesn't support Python 3.6 this PR is blocked until we do. An alternative would be to conditionally install dimod by python version but IMO it makes sense to just drop the 3.6 and rebase.

arcondello:
> Doctest fix looks good, but we should probably first drop py36 (in a separate PR?)

haha, jinx

randomir:
Just 14s... :laughing: 

image

PR #185, opened 2022-01-17
Drop support for Python 3.6

arcondello:
I think that's all of the places, we dropped it from the README.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -3,7 +3,6 @@ classifiers =
     License :: OSI Approved :: Apache Software License
     Operating System :: OS Independent
     Programming Language :: Python :: 3 :: Only
-    Programming Language :: Python :: 3.6
     Programming Language :: Python :: 3.7
     Programming Language :: Python :: 3.8
     Programming Language :: Python :: 3.9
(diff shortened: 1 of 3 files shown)


setup.cfg
```diff
@@ -46,7 +46,7 @@ install_requires =
     penaltymodel==0.16.6
     pyqubo==1.2.0
 packages = dwaveoceansdk
-python_requires = >=3.6
+python_requires = >=3.7
```
randomir, line 49: You missed it in `classifiers`.

arcondello: :flushed:

JoelPasvolsky:
https://github.com/dwavesystems/dwave-ocean-sdk/blob/master/docs/overview/install.rst

randomir (approved):

arcondello:
> https://github.com/dwavesystems/dwave-ocean-sdk/blob/master/docs/overview/install.rst

Ok, I am going to remove the specific version reference from that file.

arcondello:
> https://github.com/dwavesystems/dwave-ocean-sdk/blob/master/docs/overview/install.rst: "Python versions 3.6 and higher are supported."

Handled by https://github.com/dwavesystems/dwave-ocean-sdk/pull/186

PR #186, opened 2022-01-17
Clean up docs/conf.py a bit

arcondello:
Remove some cruft, and do some minor reorganization. Highlight is that the minimum python version is now set algorithmically from the `setup.cfg`.


docs/conf.py
```diff
@@ -1,45 +1,42 @@
...
+from dwaveoceansdk import __version__ as version
+from dwaveoceansdk import __version__ as release
+
+setup_cfg = configparser.ConfigParser()
+setup_cfg.read(os.path.join(sdk_directory, 'setup.cfg'))
+
+author = setup_cfg['metadata']['author']
+copyright = setup_cfg['metadata']['author']
+
+project = u'Ocean Documentation'
```
randomir, line 33: Unicode literal not needed in py3.

randomir (approved):
Nice job!

docs/conf.py
```diff
@@ -58,65 +55,30 @@
...
 
-# The name of the Pygments (syntax highlighting) style to use.
 pygments_style = 'sphinx'
 
-# If true, `todo` and `todoList` produce output, else they produce nothing.
 todo_include_todos = True
 
 doctest_global_setup = """
-from __future__ import print_function, division
 import operator         # Used by dwave-binarycsp
```
JoelPasvolsky, line 79: No longer needed (dwave-binarycsp examples removed)

docs/conf.py
```diff
@@ -148,179 +110,52 @@
...
-# Example configuration for intersphinx: refer to the Python standard library.
-#intersphinx_mapping = {'https://docs.python.org/': None,
-#                       'http://networkx.readthedocs.io/en/latest/': None}
-intersphinx_mapping = {'python': ('https://docs.python.org/3/', None),
+# -- Intersphinx ----------------------------------------------------------
+intersphinx_mapping = {
+    'python': ('https://docs.python.org/3/', None),
     'numpy': ('http://docs.scipy.org/doc/numpy/', None),
     'bson': ('https://api.mongodb.com/python/current/', None),
     'networkx': ('https://networkx.github.io/documentation/stable/', None),
```
JoelPasvolsky, line 152: I think this is now https://networkx.org/documentation/stable/

JoelPasvolsky (approved):
Much better. Thank you!

docs/conf.py
arcondello: looks like we're still using it somewhere, so will restore it for now

PR #190, opened 2022-02-14
Add release notes

JoelPasvolsky:
Currently adds RNs for dimod, cloud-client, and inspector.
Note for merge: requires an SDK release that includes https://github.com/dwavesystems/dimod/pull/1121, https://github.com/dwavesystems/dwave-cloud-client/pull/513, and https://github.com/dwavesystems/dwave-inspector/pull/129


arcondello:
I feel like we're moving a bit too much to "the front page", i.e. the left bar. So IMO having this nested under the relevant packages makes sense, at least until we have one unified changelog. As it stands it feels like we're subdividing by packages in two different places. But I can be convinced ether way.

JoelPasvolsky:
> I feel like we're moving a bit too much to "the front page", i.e. the left bar. So IMO having this nested under the relevant packages makes sense, at least until we have one unified changelog. As it stands it feels like we're subdividing by packages in two different places. But I can be convinced ether way.

I'm open to moving. The advantages I see for the current proposal are:
1. User can scan/search through all changes for an SDK release in place rather than under each package separately 
2. Simpler file structure: one added file rather than new files/softlinks for each package
3. Bookmark that won't change if we restructure/change our packages

randomir (commented):
Looks good, but I would prefer we keep package release notes under each package, and Ocean release notes under the main "Release notes" (on the sidebar link).

Not sure there's an easy way to collect them from the SDK release page, as that's currently the only place we compile them for all packages.

Btw, dwave-inspector also uses reno.

arcondello:
> Seems that's the preference, so I'll move and add for inspector. A point I forgot to note above is that in the near term only three of the packages have RNs so to find them under packages, users would be clicking through and mostly coming up empty (3/15).

Honestly I think that's fine.

JoelPasvolsky:
So for most of our repos, placing release notes under packages is straightforward. However, repos such as dwave-cloud-client have different paths from the ``docs`` directory to the repo in the SDK from the submodule.

To overcome that, I considered these methods:

1. Cleanest would be using the ``ifconfig`` directive to set the ``reporoot`` with either the substitution directive (preferred) or the entire ``release-notes`` directive. Neither works because the ``ifconfig`` was designed for textual content

2. Adding a symlink to the top-level directory of dwave-cloud-client that links `dwave-cloud-client` to `dwave/cloud` and using that for the ``reporoot``. I worry it might cause problems, so before I do that I'm running it by you two

3. Setting ``reporoot`` just to the SDK's path. This works also for the local build because  https://github.com/openstack/reno/blob/cd31b907c380635dd8e9c61995a1c0826d4a1137/reno/sphinxext.py#L59 but I don't want to rely on that, which might change tomorrow.

4. Putting both paths. Will build and render okay but will give quite a few build errors for duplication.

5.  Setting it correctly for the SDK and not worrying anymore about local builds (for cloud client)

Any other ideas? Or should we go with 5?

arcondello:
I am fine with 5. The local builds are mostly for testing anyway.

randomir:
Yup, 5 is fine.

PR #221, opened 2022-07-29
Update packages for 6.1.0

arcondello:


arcondello:
Current test failures seem to be from an interaction between https://github.com/dwavesystems/dwave-neal/pull/91 and dwave-hybrid. Can look into it next week.

randomir:
@arcondello, Test failures stem from a backward-compatibility breaking change in [dwave-neal#91](https://github.com/dwavesystems/dwave-neal/pull/91) -- `num_sweeps` now has to be positive.

And in dwave-hybrid (tests) we set `num_sweeps=0` to verify initial state is correctly propagated through the SA runnable.

We can rework that test, but IMO, it's nice to have a "NOP" SA/sampler input. OTOH, supporting that in the current Neal code does not seem straightforward.

arcondello:
> Looks good, it's just not clear to me why are we adding docs manually here, instead of using Neal's docs (with deprecation note added there)?

We talked about this offline, but for posterity it could be done either way. I slightly prefer this one because then we use the version number of the sdk in the deprecation warning. Also makes the existing dwave-neal repo a bit smaller/tidier.

arcondello:
I have the wrong version number in the deprecation notice everywhere. I just wanted it to be easy to find.

docs/docs_tabu/intro.rst
```diff
@@ -0,0 +1,28 @@
...
+‘sample_qubo’ and ‘sample_ising’ methods as well as the generic BQM sampler method.
+
+The :class:`~tabu.TabuSampler` sampler implements the `MST2 multistart tabu search algorithm
+<https://link.springer.com/article/10.1023/B:ANOR.0000039522.58036.68>`_
+for quadratic unconstrained binary optimization (QUBO) problems
+with a :std:doc:`dimod <oceandocs:docs_dimod/sdk_index>` Python wrapper.
+
+For a description of the tabu search algorithm, see `tabu search <https://en.wikipedia.org/wiki/Tabu_search>`_\ .
+
+Example
```
JoelPasvolsky, line 23: This need to be deleted, it shows an empty Example section

docs/licenses.rst
```diff
@@ -10,12 +10,9 @@ Below are the open-source licenses for the SDK and its tools:
...
    licenses/cloud
-   licenses/greedy
    licenses/hybrid
-   licenses/neal
    licenses/dnx
    licenses/system
-   licenses/tabu
    licenses/minorminer
    licenses/penaltymodel
    licenses/preprocessing
```
JoelPasvolsky, line 18: Needs a `licenses/samplers` line and consider for [dwave-samplers](https://github.com/dwavesystems/dwave-samplers/tree/main/docs) using an `sdk_index.rst` file for consistency (or removing in other repos, now we're not doing per-repo builds)

docs/packages.rst
```diff
@@ -18,10 +18,11 @@ Packages
   docs_inspector/sdk_index
   docs_neal/sdk_index
   docs_dnx/sdk_index
+  docs_preprocessing/sdk_index
+  docs_samplers/index
```
JoelPasvolsky, line 22: Consider an sdk_index (even though we're not building for the repo, be good to keep consistent)

JoelPasvolsky (commented):
Looks good

Open an issue to update places like [stack](https://docs.ocean.dwavesys.com/en/stable/overview/stack.html), [cpu](https://docs.ocean.dwavesys.com/en/stable/overview/cpu.html), [pp_greedy](https://docs.ocean.dwavesys.com/en/stable/examples/pp_greedy.html)

docs/packages.rst
arcondello: I think we should be updating the other ones to drop `sdk_index` and just use `index`

arcondello:
> I'm a bit confused why you removed docs from greedy/neal/tabu, only to add them here (and remove submodules).

We could have done it either way. [As we talked about before](https://github.com/dwavesystems/dwave-ocean-sdk/pull/221#issuecomment-1213329682), the main reason was to keep the deprecation warnings easy to update and able to use the sdk version.

randomir (approved):

JoelPasvolsky (commented):
Add also to [licenses](https://docs.ocean.dwavesys.com/en/stable/licenses.html), 2nd to last

arcondello:
> Add also to [licenses](https://docs.ocean.dwavesys.com/en/stable/licenses.html), 2nd to last

We don't do pyqubo? And we haven't brought it in as a submodule so there is no file to link in the same way

setup.cfg
```diff
@@ -44,3 +45,4 @@ python_requires = >=3.7
 
 [options.extras_require]
 all =
+    dwave-gate==0.1.0
```
randomir, line 48: I was thinking something more specific than `all`, since we don't include drivers and the inspector with `all`.. Perhaps `gate`? Have no strong opinion on this one.

arcondello: I also have no strong opinion. `all` seemed safe but happy to do it in others as well.

randomir (approved):

PR #229, opened 2022-09-23
Add GTM

JoelPasvolsky:


arcondello (approved):
LGTM, though, to be fair, I have no idea what this means :laughing:

dwave-pmilesi (approved):
Works for me !

randomir (commented):
Shouldn't old universal analytics be now [implemented through GTM](https://support.google.com/tagmanager/answer/6107124?hl=en)?

dwave-pmilesi:
> Shouldn't old universal analytics be now [implemented through GTM](https://support.google.com/tagmanager/answer/6107124?hl=en)?

They could but this is just temporary until we switch completely to GA4 and GTM, and then we'll remove UA code (UA will be retired next year). By using 2 snippets, 2 separate dataLayer objects are used to push data to the properties, and so prevent any possibility of duplicated data.

randomir:
Got it, @dwave-pmilesi! In that case, LGTM! :slightly_smiling_face:

randomir (approved):

PR #236, opened 2022-11-09
Add a presolve example

JoelPasvolsky:


docs/examples/pp_cqm_presolve.rst
```diff
@@ -0,0 +1,93 @@
...
+
+.. testcode::
+
+    import dimod
+
+    j = dimod.Integer("j")
+    cqm = dimod.ConstrainedQuadraticModel()
+    cqm.add_constraint(j <= 5, "Maximum j")
+
+Clearly, the global optimum for this CQM occurs for the default value of the lower
```
arcondello, line 57: That's not true, any of the values 0, 1, 2, 3, 4, 5 are optimal for this problem.

docs/examples/pp_cqm_presolve.rst
```diff
@@ -0,0 +1,93 @@
...
+bound of ``j``.
+
+>>> cqm.lower_bound("j")
```
arcondello, line 60: This line seems a bit inscrutable to me. Are you trying to show that presolve will adjust the bounds?

docs/examples/pp_cqm_presolve.rst
```diff
@@ -0,0 +1,93 @@
...
+The dimod :class:`~dimod.reference.samplers.ExactCQMSolver` test solver is
+capable of solving this very simple CQM.
+
+>>> sampleset = dimod.ExactCQMSolver().sample_cqm(reduced_cqm)
+>>> print(sampleset.first)
+Sample(sample={0: 0}, energy=0.0, num_occurrences=1, is_satisfied=array([], dtype=bool), is_feasible=True)
+
+View the solution as assignments of the original CQM's variables: 
+
+>>> presolve.restore_samples(sampleset.first.sample)
```
arcondello, line 92: I would follow this step with `sampleset = dimod.SampleSet.from_samples_cqm(samples, cqm)` to restore the "full" sample set, rather than just the samples as seen here.

docs/examples/pp_cqm_presolve.rst
randomir: :100:

docs/examples/pp_cqm_presolve.rst
```diff
@@ -0,0 +1,93 @@
...
+The CQM created below has a constraint requiring that integer variable ``j``
+not exceed 5.
+
+.. testcode::
+
+    import dimod
+
+    j = dimod.Integer("j")
+    cqm = dimod.ConstrainedQuadraticModel()
+    cqm.add_constraint(j <= 5, "Maximum j")
```
randomir, line 55: I would add an objective here, otherwise presolving is pretty pointless. How about:
```suggestion
    j = dimod.Integer("j")
    cqm = dimod.ConstrainedQuadraticModel()
    cqm.set_objective(-j)
    cqm.add_constraint(j <= 5, "Maximum j")
```

docs/examples/pp_cqm_presolve.rst
JoelPasvolsky: Forgot, no objective 🤦

docs/examples/pp_cqm_presolve.rst
JoelPasvolsky: See if [the last two commands](https://github.com/dwavesystems/dwave-ocean-sdk/pull/236/commits/55597fae6d2860505bc305ca53ccd2e2dbd6a2b0) are what you were thinking of

docs/examples/pp_cqm_presolve.rst
JoelPasvolsky: Actually, an objective with `+j` works nicer for this example

arcondello (commented):
I am happy with the example, but I think this should be a module-level docstring in dwave-preprocessing rather than a full "example" in the sdk docs. We don't yet know if we want to encourage users to always run presolve locally, so IMO until we have more data it makes sense to keep it in the more easily maintained place.

docs/examples/pp_cqm_presolve.rst
```diff
@@ -0,0 +1,107 @@
...
+
+The purpose of this example is to show how to use Ocean's
+:class:`~dwave.preprocessing.presolve.Presolver` class.
+
+Example Requirements
+====================
+
+To run the code in this example, the following is required.
+
+* The requisite information for problem submission through SAPI, as described in :ref:`sapi_access`.
```
arcondello, line 23: I believe this example does not actually use any online resources

JoelPasvolsky:
> should be a module-level docstring in dwave-preprocessing

See [this commit](https://github.com/dwavesystems/dwave-preprocessing/pull/60/commits/c39f2c73861bbaa2a94c7781f11532ca26da63db)

docs/examples/pp_cqm_presolve.rst
randomir: Why do you think? IMO, `-j` is nicer as it uses the visible (just defined) upper bound -- by converting it to offset.

JoelPasvolsky: The example is to show the flow of actions so simpler is better, and with `+j` the lower bound is zero versus a high number, they don't have to do the switch in their head from minimizing a negative to maximizing a positive, and don't need to understand offset.

arcondello:
Replaced by https://github.com/dwavesystems/dwave-preprocessing/pull/60

PR #237, opened 2022-11-10
Make package names doc links

JoelPasvolsky:
See if this helps prevent the initial misclick through to package docs

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -31,7 +31,7 @@ The Ocean SDK includes the :ref:`dwave_cli` and the following packages:
 
 .. packages-start-marker
 
-.. dropdown::  **dimod** --- Quadratic models: BQM, DQM. \
+.. dropdown::  `dimod <https://docs.ocean.dwavesys.com/en/stable/docs_dimod/sdk_index.html>`_ --- Quadratic models: BQM, DQM. \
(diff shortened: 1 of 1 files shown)


randomir (commented):
Can you put `<docs>` pill next to each docs link (package name), and add a GitHub icon next to the `<code>` pill?

JoelPasvolsky:
> Can you put `<docs>` pill next to each docs link (package name), and add a GitHub icon next to the `<code>` pill?

I can shift the <docs> pill to the left, which will keep them all aligned:

image

I can add the GitHub cat but that (a) I don't know how to float right (b) haven't put effort into finding out because I think having three pills/icons is a bit much and will confuse users that clicking a pill takes them somewhere but clicking the cat opens the panel

JoelPasvolsky:
> Another idea, how about adding a TOC (of depth 1) in each collapsible package box? So I can jump to "reference documentation" that I'm interested in (99.99% times).

I can do something like this: 
image

randomir:
@JoelPasvolsky, I like how the `<docs>` pill looks on the left.

Re: github icon -- I meant we group it with the `<code>` pill, e.g. `<(octocat) code>`, or `(octocat) <code>`, so it's clear the link will take you to GitHub source...

JoelPasvolsky:
> Re: github icon -- I meant we group it with the `<code>` pill, e.g. `<(octocat) code>`, 

I had a look at doing that but don't see that bootstrap pills support an icon

PR #247, opened 2023-01-26
Add dwave-gate docs

thisac:
Adds the dwave-gate docs to the SDK.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -124,11 +124,7 @@ jobs:
                   fi
                   python setup.py build_ext --inplace 2>/dev/null
                 fi
-                if [[ -f "requirements_dev.txt" ]] && grep -q pytest "requirements_dev.txt"; then
-                  python -m pytest tests/
-                else
-                  python -m unittest discover
-                fi
+                python -m pytest tests
                 cd ../
               fi
             done
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -102,13 +102,16 @@ jobs:
               if [[ -f "$subdir"/test_requirements.txt ]]; then
                 pip install -r "$subdir"/test_requirements.txt
               fi
+              if [[ -f "$subdir"/requirements_dev.txt ]]; then
+                pip install -r "$subdir"/requirements_dev.txt
+              fi
               done
-            pip install cython wheel
+            pip install cython wheel cgen
             pip install .[all]
             dwave install --all --yes
 
       - run:
-          name: run individual package unittests
+          name: run individual package unit tests
           command: |
             . env/bin/activate
             for subdir in */; do
@@ -116,9 +119,16 @@ jobs:
                 cd "$subdir"
                 pwd
                 if [[ -f "setup.py" ]]; then
+                  if [[ "$subdir" == "dwave-gate/" ]]; then
+                    python dwave/gate/simulator/operation_generation.py
+                  fi
                   python setup.py build_ext --inplace 2>/dev/null
                 fi
-                python -m unittest discover
+                if [[ -f "requirements_dev.txt" ]] && grep -q pytest "requirements_dev.txt"; then
+                  python -m pytest tests/
+                else
+                  python -m unittest discover
+                fi
                 cd ../
               fi
             done
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -104,7 +104,7 @@ jobs:
               fi
               done
             pip install cython wheel
-            pip install .
+            pip install .[all]
             dwave install --all --yes
 
       - run:
@@ -141,7 +141,7 @@ jobs:
           command: |
             . env/bin/activate
             python --version
-            pip install .
+            pip install .[all]
 
       - run:
           name: install docs requirements
@@ -185,7 +185,7 @@ jobs:
           name: verify version matches tag
           command: |
             . env/bin/activate
-            pip install .
+            pip install .[all]
             echo $CIRCLE_TAG
             [[ "$(pip show dwave-ocean-sdk 2>/dev/null | grep Version)" == "Version: $CIRCLE_TAG" ]]
 
@@ -229,7 +229,7 @@ jobs:
             python3 --version
             python3 -m venv env
             . env/bin/activate
-            pip install .
+            pip install .[all]
 
       # we have to mangle the pyproject.toml becuase otherwise black duck
       # decides that we must be using poetry and fails. Luckily we don't
--- a/.gitmodules
+++ b/.gitmodules
@@ -31,3 +31,6 @@
 [submodule "dwave-samplers"]
 	path = dwave-samplers
 	url = https://github.com/dwavesystems/dwave-samplers.git
+[submodule "dwave-gate"]
+	path = dwave-gate
+	url = git@github.com:dwavesystems/dwave-gate.git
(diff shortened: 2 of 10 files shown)


setup.cfg
```diff
@@ -29,6 +29,7 @@ install_requires =
     dimod==0.12.3
     dwave-cloud-client==0.10.4
     dwave-greedy==0.3.0
+    dwave-gate==0.1.0
```
arcondello, line 32: I don't think we want to add it to the "main packages" yet?

thisac: Ah, good point.

arcondello:
Looks like something is going wrong
image
It's also not showing up in the build

thisac:
@arcondello Forgot to commit the dwave-gate submodule. Should hopefully work now.

arcondello:
```
**********************************************************************
File "docs_gate/README.rst", line 22, in default
Failed example:
    from dwave.gate.simulator import simulate
Exception raised:
    Traceback (most recent call last):
      File "/home/circleci/.pyenv/versions/3.7.16/lib/python3.7/doctest.py", line 1337, in __run
        compileflags, 1), test.globs)
      File "<doctest default[0]>", line 1, in <module>
        from dwave.gate.simulator import simulate
    ModuleNotFoundError: No module named 'dwave.gate'
**********************************************************************
File "docs_gate/README.rst", line 23, in default
Failed example:
    simulate(circuit)
Exception raised:
    Traceback (most recent call last):
      File "/home/circleci/.pyenv/versions/3.7.16/lib/python3.7/doctest.py", line 1337, in __run
        compileflags, 1), test.globs)
      File "<doctest default[1]>", line 1, in <module>
        simulate(circuit)
    NameError: name 'simulate' is not defined
**********************************************************************
```

need to add `[all]` to https://github.com/dwavesystems/dwave-ocean-sdk/blob/ebc494e2da2d23ab02af5f1af7eeeaf023ab570e/.circleci/config.yml#L144

docs/conf.py
```diff
@@ -190,6 +190,7 @@ def setup(app):
 github_map = {'dwavebinarycsp': 'dwavebinarycsp',
               'cloud': 'dwave-cloud-client',
               'dimod':  'dimod',
+              'gate': 'dwave-gate',
```
arcondello, line 193: I think the error is related to this line.

thisac: Maybe I'm missing something obvious here, but I don't see why that would be the case? 

All the generated URLs seem to be valid. 🤔

.gitmodules
```diff
@@ -31,3 +31,6 @@
 [submodule "dwave-samplers"]
 	path = dwave-samplers
 	url = https://github.com/dwavesystems/dwave-samplers.git
+[submodule "dwave-gate"]
+	path = dwave-gate
+	url = git@github.com:dwavesystems/dwave-gate.git
```
randomir, line 36: Should use https, so that it can be checked out anonymously:
```suggestion
	url = https://github.com/dwavesystems/dwave-gate.git
```

docs/conf.py
JoelPasvolsky: The failure is due to [this removal](https://github.com/dwavesystems/dwave-ocean-sdk/pull/247#discussion_r1088386301): the ``sphinx.ext.linkcode`` needs the version, which is calculated as ``pkg_resources.get_distribution('dwave-ocean-sdk').requires()``. 

You can get the build to pass by adding it manually as ``versions['dwave-gate'] = '0.1.0'``. Until you make ``dwave-gate`` a requirement of the SDK, you should set it to automatically add the version of ``dwave-gate``.

.circleci/config.yml
```diff
@@ -102,23 +102,33 @@ jobs:
...
+                    python dwave/gate/simulator/operation_generation.py
+                  fi
                   python setup.py build_ext --inplace 2>/dev/null
                 fi
-                python -m unittest discover
+                if [[ -f "requirements_dev.txt" ]] && grep -q pytest "requirements_dev.txt"; then
+                  python -m pytest tests/
+                else
+                  python -m unittest discover
+                fi
```
thisac, line 131: Perhaps not the most elegant solution, but it works. Could also just use pytest for dwave-gate (and unittest for all others), or check for pytest in e.g., `pyproject.toml` and add it as an install-requires line there since this relies on there being a `requirements_dev.txt` file with `pytest` added as a dependency.

thisac:
@arcondello @randomir @JoelPasvolsky This should be good for a (final) review now!

.circleci/config.yml
randomir, line 131: Why not just:
```suggestion
                python -m pytest tests
```

Pytest handles Python unittests OOTB. We would just need to always install `pytest` (few lines above).

.circleci/config.yml
```diff
@@ -102,23 +102,33 @@ jobs:
               if [[ -f "$subdir"/test_requirements.txt ]]; then
                 pip install -r "$subdir"/test_requirements.txt
               fi
+              if [[ -f "$subdir"/requirements_dev.txt ]]; then
+                pip install -r "$subdir"/requirements_dev.txt
+              fi
```
randomir, line 107: Are you introducing a new convention here?

All* Ocean packages have one `requirements.txt` in root and then one in `tests/` and possibly `docs/`. Your `requirements_dev.txt` would be `tests/requirements.txt` in our current convention.

*) all except minorminer. Not sure why we deviated with `test_requirements.txt` there. :thinking:

thisac: I guess I am, although not intentionally. This was never questioned previously, but I'd be happy to follow conventions as much as possible. Can switch to `tests/requirements.txt` if that's the preferred one.

.circleci/config.yml
thisac: Ah, didn't realize that. Much simpler then!

thisac: @randomir Tests now fail with `AttributeError: 'TestMarginals' object has no attribute 'bqm'` in `dwave-samplers`. 😱

arcondello: Different test discovery patterns. `pytest` does by class name but `unittest` does by inheritance from `unittest.TestCase`

arcondello: To be fair, it's an odd pattern in dwave-samplers, not sure why I wrote it that way (at least I think it was me). Probably should get a refactor for clarity at some point https://github.com/dwavesystems/dwave-samplers/blob/a16f7b9c929632a1807347a486bff43950d45d64/tests/test_tree_sampler.py#L110-L281

arcondello: I'll just make a PR to dwave-samplers today. Should be quick and supporting other testing frameworks like `pytest` is a nice feature if we can manage it.

randomir: Yeah, renaming `TestMarginals` mixin should do the trick.

arcondello: With https://github.com/dwavesystems/dwave-samplers/pull/49 merged, I think all you need to do is update the commit of dwave-samplers to fix the tests.

thisac: @arcondello Seems to be a similar issue [here ](https://github.com/dwavesystems/dwave-system/blob/master/tests/test_embedding_chain_breaks.py#L64)with dwave-system.

thisac: @arcondello @randomir `test_individual_pkgs` is now passing (along with all other checks). Should be ready for another review/approval. Thanks for the help with getting this in!

arcondello (approved):
LGTM

randomir (approved):

PR #250, opened 2023-02-17
Fix linkcheck failure

JoelPasvolsky:
``curl`` is still rejected this morning from this site


arcondello (approved):

randomir (commented):
Looks like they've built a deliberate protection against scrapers.

They require `User-Agent`, encrypted ID `Cookie` set on first request, and JavaScript..

Sad.

PR #252, opened 2023-02-17
Fix RTD YAML file for dwave-gate

JoelPasvolsky:
Since the [dwave-gate docs PR](https://github.com/dwavesystems/dwave-ocean-sdk/pull/247), the doc build is failing with ``IndexError: list index out of range`` error on line [this line of conf.py](https://github.com/dwavesystems/dwave-ocean-sdk/blob/e62df3842f80bbc08cc4a8f000927ee3e8426ef9/docs/conf.py#L209). I think this should fix it.

The ``conf.py`` file might benefit from also being updated to use [importlib.resources, importlib.metadata](https://setuptools.pypa.io/en/latest/pkg_resources.html) instead of ``pkg_resources`` at some point, as noted therr.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -12,6 +12,7 @@ python:
    install:
       - method: pip
         path: .
+        extra_requirements: [all]
       - requirements: docs/requirements.txt
         
 


arcondello:
Reviewer to edited characters ratio seems a bit high :laughing: . Removing myself

thisac (approved):
Shouldn't the `requires(extras=['all'])` addition [here ](https://github.com/dwavesystems/dwave-ocean-sdk/blob/e62df3842f80bbc08cc4a8f000927ee3e8426ef9/docs/conf.py#L208) be enough? Looks like a reasonable change, though, so hoping that this fixes it.

JoelPasvolsky:
> Shouldn't the `requires(extras=['all'])` addition [here ](https://github.com/dwavesystems/dwave-ocean-sdk/blob/e62df3842f80bbc08cc4a8f000927ee3e8426ef9/docs/conf.py#L208) be enough? Looks like a reasonable change, though, so hoping that this fixes it.

That correctly takes all the packages required from the SDK but RTD has not installed the `dwave-gate` package locally as part of its `pip install` so it doesn't locate that package in the venv's ``site-packages``

randomir (approved):
LGTM!

JoelPasvolsky:
dwave-gate showing now: https://docs.ocean.dwavesys.com/en/latest/docs_gate/sdk_index.html

PR #253, opened 2023-04-25
Update AND example to use newer QPU

JoelPasvolsky:
Updated to run on Advantage rather than the older D-Wave 2000Q. I kept the same example idea of an AND gate because we have very many references to this example.


docs/examples/and.rst
```diff
@@ -42,18 +42,18 @@ formulation of a BQM for a Boolean gate in detail. This example briefly repeats
...
 * Qubit coupling strengths: control the degree to which two qubits tend to the same state.
 
 The biases and couplings define an energy landscape, and the D-Wave quantum computer seeks
-the minimum energy of that landscape. Once you express your problem in a formulation\ [#]_
+the minimum energy of that landscape. Once you express your problem in a formulation\ [1]_
 such that desired outcomes have low energy values and undesired outcomes high energy values,
 the D-Wave system solves your problem by finding the low-energy states.
 
-.. [#] This formulation, called an :term:`objective function`, corresponds to the :term:`Ising`
+.. [1] This formulation, called an :term:`objective function`, corresponds to the :term:`Ising`
```
arcondello, line 56: Why switch to explicit numbering? Won't that make it harder to maintain in the future? Also inconsistent with the citation you added lower down.

docs/examples/and.rst
```diff
@@ -131,16 +131,17 @@ The line of code below sets the QUBO coefficients for this AND gate.
...
+Solve the Problem by Sampling (Automated Minor-Embedding)
+=========================================================
+
+For reference, first solve with the same steps used in the :ref:`not` example.
 
 Again use sampler :class:`~dwave.system.samplers.DWaveSampler` from Ocean software's
 :doc:`dwave-system </docs_system/sdk_index>` and
 its :class:`~dwave.system.composites.EmbeddingComposite` composite to :term:`minor-embed`
-our unstructured problem (variables x1, x2, and z) on the sampler's graph structure (the
+the unstructured problem (variables x1, x2, and z) on the sampler's graph structure (the
```
arcondello, line 144:
```suggestion
the unstructured problem (variables :math:`x1`, :math`x2`, and :math`z`) on the sampler's graph structure (the
```

arcondello (approved):

docs/examples/and.rst
randomir: There's [this](https://docutils.sourceforge.io/docs/user/rst/quickref.html#footnotes):

```suggestion
.. [#ising] This formulation, called an :term:`objective function`, corresponds to the :term:`Ising`
```

docs/examples/and.rst
```diff
@@ -42,18 +42,18 @@ formulation of a BQM for a Boolean gate in detail. This example briefly repeats
...
-a D-Wave 2000Q has up to 2048 qubits connected in a :term:`Chimera` topology. Programming it
+an Advantage QPU has over 5000 qubits connected in a :term:`Pegasus` topology. Programming it
 consists mostly of setting two inputs:
 
 * Qubit bias weights: control the degree to which a qubit tends to a particular state.
 * Qubit coupling strengths: control the degree to which two qubits tend to the same state.
 
 The biases and couplings define an energy landscape, and the D-Wave quantum computer seeks
-the minimum energy of that landscape. Once you express your problem in a formulation\ [#]_
+the minimum energy of that landscape. Once you express your problem in a formulation\ [1]_
```
randomir, line 52:
```suggestion
the minimum energy of that landscape. Once you express your problem in a formulation\ [#ising]_
```
(the second half of the comment below; GH doesn't allow suggested change over two diff snippets)

docs/examples/and.rst
```diff
@@ -194,62 +195,87 @@ following coefficients:
...
+    :align: center
+    :scale: 50 %
+
+    A NOT gate minor embedded onto an Advantage QPU. Variables :math:`x,z` 
+    (left) are minor-embedded to qubits ``30`` and ``31`` (right). This 
+    visualization is produced by Ocean's :std:doc:`docs_inspector/sdk_index` tool.
+    
+Notice that coupling strength is :code:`BIAS 0.5` above: as explained in the 
+:std:doc:`system documentation <sysdocs_gettingstarted:doc_getting_started>`,
+the QPU can be viewed as minimizing the Ising model of [1_] above, with linear
```
randomir, line 230: autonumbered footnot can be used here as well

docs/examples/and.rst
```diff
@@ -264,113 +290,174 @@ From NOT to AND: an Important Difference
...
+    (left) are minor-embedded to qubits ``30``, ``31`` and chain ``[45, 46]`` (right). 
+
+
+Chain Strength 
+--------------
+
+For illustrative purposes, the following code purposely weakens the
+:ref:`chain strength <concepts__chain_strength>` (strength of the coupling 
+between qubits ``45`` and ``46``, which represent variable :math:`z`). By 
+explicitly setting chain strength to a low value of 0.25, the two qubits are
```
randomir, line 427: It would be useful to link to some reference that explains why 0.25 is "low value" (not sure where exactly would that be, though).

randomir (approved):
LGTM!

docs/examples/and.rst
JoelPasvolsky: Required some modification of the text and your naming but looks nice now: 
image
image

docs/examples/and.rst
JoelPasvolsky: Added:
image

randomir: That's great, thanks.

PR #254, opened 2023-04-26
Update concepts to use newer QPU

JoelPasvolsky:
Updated Concept sections to run on Advantage rather than the older D-Wave 2000Q. Also added a couple of glossary terms requested by training and fixed some barring caused by inconsistent indentation.


docs/concepts/solutions.rst
```diff
@@ -28,43 +34,47 @@ SampleSets
...
->>> bqm = dimod.BQM({}, {('s0', 's1'): -1, ('s0', 's2'): -1, ('s1', 's2'): 1}, 0, dimod.Vartype.SPIN)  
->>> sampler = EmbeddingComposite(DWaveSampler(solver={'qpu': True}))     
+>>> import dimod 
+>>> from dwave.system import DWaveSampler, EmbeddingComposite
+...
+>>> bqm = dimod.BinaryQuadraticModel(
+...           {}, 
+...           {('s0', 's1'): -1, ('s0', 's2'): -1, ('s1', 's2'): 1}, 
+...           0, 
+...           dimod.Vartype.SPIN)
```
randomir, line 59: Why keep it simple for this concepts page, like:
```suggestion
>>> bqm = dimod.BQM.from_ising({}, {('s0', 's1'): -1, ('s0', 's2'): -1, ('s1', 's2'): 1})
```

or even :slightly_smiling_face: 
```suggestion
>>> bqm = dimod.BQM.from_ising({}, {'ab': -1, 'bc': 1, 'ca': -1})
```

randomir (approved):
LGTM

docs/concepts/solutions.rst
JoelPasvolsky: I think we defined that in our examples we use the full ``dimod.BinaryQuadraticModel`` and not the short form ``dimod.BQM`` and use the ``('a', 'b'): 1`` form and not the short ``'ab': 1``. But I like the ``.from_ising``, and will switch to that if you edit one of the above

randomir: Like this?
```suggestion
>>> bqm = dimod.BinaryQuadraticModel.from_ising({}, {('s0', 's1'): -1, ('s0', 's2'): -1, ('s1', 's2'): 1})
```

The line is too long, and overall it's too verbose. IMO, intro examples should be short and sweet.

randomir: But sure, this is not the place to deviate from the rest of our docs if we are consistent! Just a food for thought. :slightly_smiling_face:

arcondello: I've shown the `{'ab': -1, 'bc': 1, 'ca': -1}` format before and immediately gotten the question "how do you encode integers?". So I think we should be explicit and use the clearest and most adaptable Python possible.

JoelPasvolsky: Is it worth revisiting the ``dimod.BinaryQuadraticModel`` versus ``dimod.BQM`` guidance? I think we've cultivated the ``bqm`` pretty thoroughly by now

arcondello: I personally prefer the more explicit long form for docs. I don't think `BQM` is a common enough acronym (yet). Though of course, by that logic, we should be using `.from_quadratic_unconstrained_binary_optimization()`  rather than `.from_qubo()` so...

arcondello: If we want to do short and sweet, IMO we should just go straight to symbolic
```python
a, b, c = dimod.Binaries(['a', 'b', 'c'])
model = -1*a*b + b*c - c*a
```

randomir: Yes, that's nice. Although it adds another layer to unwrap, at least until we switch to symbolic models all over.

JoelPasvolsky: Thanks, that's very nice.

One interesting thing I notice is the signs for zero values for a BQM instantiated this way:

```
>>> s0, s1, s2 = dimod.Spins(['s0', 's1', 's2'])
>>> bqm = s1*s2 - s0*s1 - s0*s2
>>> bqm.linear
{'s1': -0.0, 's2': -0.0, 's0': -0.0}
```

We don't get those minus signs when we do ``dimod.BQM({}, {('s0', 's1'): -1, ('s0', 's2'): -1, ('s1', 's2'): 1}, 'SPIN').

arcondello: It's just an artifact of the math, `-0.0 == 0.0` :shrug:

arcondello (approved):

Issue #89, opened 2020-09-09
Update CONTRIBUTING


arcondello:

JoelPasvolsky:
Need to link to clang file in minorminer or to bring over that file.

arcondello:
While we're making changes to CONTRIBUTING, it also currently says that all code should be python 2 and 3 compatible.

PR #259, opened 2023-06-30
Refresh the contributing guide

arcondello:
Closes https://github.com/dwavesystems/dwave-ocean-sdk/issues/89

Obviously there is much here that can be debated. I tried to err on the side of clarifying existing conventions rather than adding new ones.

The biggest change is in the C++ section. In that case I included the `.clang-format` file we're currently using in [dimod](https://github.com/dwavesystems/dimod/blob/main/.clang-format) and [dwave-preprocessing](https://github.com/dwavesystems/dwave-preprocessing/blob/main/.clang-format). [minorminer](https://github.com/dwavesystems/minorminer/blob/main/.clang-format) uses a similar one but uses a 120 character limit.
I have a weak preference for the 100 character limit, but am fine either way.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -59,7 +59,7 @@ Coding Conventions
 * Functions should do one thing.
 * Early pull requests and code reviews.
 * Early architecting/design. Code reviews can happen before any code has been written.
-* Use a consistent character width of 120.
+* Use a consistent character width limit of 120.
 * Use 4 spaces instead of tabs.
 * End all files with a newline.
 
@@ -103,10 +103,20 @@ Documentation and Comments
 Python
 ------
 
+Python Version
+~~~~~~~~~~~~~~
+
+Code should be compatible with all `supported versions <https://devguide.python.org/versions/#supported-versions>`_
+of Python.
+
 pep8
 ~~~~
 
 As a baseline, follow the `pep8 <https://www.python.org/dev/peps/pep-0008/>`_ style guide for python.
+
+We do make one exception for `maximum line length <https://peps.python.org/pep-0008/#maximum-line-length>`_.
+In Ocean we maintain a hard limit of 120 characters. It is encouraged but not required to keep lines of code
+to 100 characters or fewer and docstrings/comments to 72 characters or fewer.
 
 Documentation
 ~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


arcondello:
CI errors seem to be https://github.com/dwavesystems/dwave-cloud-client/issues/553.

CONTRIBUTING.rst
```diff
@@ -131,18 +141,38 @@ C++
...
+See `PyPA's docs <https://cibuildwheel.readthedocs.io/en/stable/cpp_standards/>`_ on C++ standards for a discussion
+on how this interacts with Python wheels.
 
 Format
 ~~~~~~
 
 * Our format is based on `Google C++ style guide <https://google.github.io/styleguide/cppguide.html>`_ with some exceptions:
 
-  - Column width is limited to 120 characters. Best effort should be made to keep to 80 characters, but up to 120 can be used for clarity.
+  - Column width is limited to 100 characters. Best effort should be made to keep to 80 characters, but up to 120 can be used for clarity.
```
arcondello, line 154: I see there is a second 120 that I missed.

randomir (approved):
lgtm!

CONTRIBUTING.rst
```diff
@@ -59,7 +59,7 @@ Coding Conventions
 * Functions should do one thing.
 * Early pull requests and code reviews.
 * Early architecting/design. Code reviews can happen before any code has been written.
-* Use a consistent character width of 120.
+* Use a consistent character width limit of 120.
```
thisac, line 62: Minor, but this sounds strange to me. Shouldn't it be either "character limit" or "line width"?

thisac (approved):
LGTM. Could be slightly clearer on character limits for Python by setting it to 100 (as is recommended later in the doc) instead of 120, even though Ocean uses 120, but I don't really have a strong opinion here.

CONTRIBUTING.rst
```diff
@@ -103,11 +103,21 @@ Documentation and Comments
...
+of Python.
+
 pep8
 ~~~~
 
 As a baseline, follow the `pep8 <https://www.python.org/dev/peps/pep-0008/>`_ style guide for python.
 
+We do make one exception for `maximum line length <https://peps.python.org/pep-0008/#maximum-line-length>`_.
+In Ocean we maintain a hard limit of 120 characters. It is encouraged but not required to keep lines of code
+to 100 characters or fewer and docstrings/comments to 72 characters or fewer.
```
JoelPasvolsky, line 119: "docstrings/comments to 72 characters" and a few lines above it we give [an example](https://github.com/dwavesystems/dwave-ocean-sdk/blob/6b4d6cd53621e2b5f662064ba1a94dd14f93db57/CONTRIBUTING.rst?plain=1#L91) with a comment of length 96 characters

CONTRIBUTING.rst
JoelPasvolsky, line 62: I think people will see this and will use 120 as their standard length. We should say here our recommended 100 (with an asterisk) and let the  pep-8 section then present the hard limit of 120

JoelPasvolsky (approved):
Thank you

PR #261, opened 2023-08-10
Drop support for Python 3.7

randomir:
- drop py37
- upgrade xcode to fix CircleCI macos builds on the medium resource class
- upper-bound cython (long term we'll have to upgrade to cython 3, which is finally out)

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -4,10 +4,10 @@ environment:
 
     # https://www.appveyor.com/docs/build-environment/#python
 
-    - PYTHON: "C:\\Python37-x64"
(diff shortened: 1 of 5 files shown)


thisac (approved):
Very minor, but could update the example Python version in[ the bug report issue template](https://github.com/randomir/dwave-ocean-sdk/blob/04883f87092a80ba13cbd9aa50e4b2abb93e7cbe/.github/ISSUE_TEMPLATE/bug_report.md). 

Otherwise, LGTM!

randomir:
Nice catch, will do.

PR #274, opened 2023-11-09
WIP: Update doc-build tools to support new ``pydata-sphinx-theme`` & switch from obsolete ``sphinx-panels``

JoelPasvolsky:
I'll open this WIP now but will continue after my PTO. I've fixed the worst of the backwards incompatibilities but still need to 

- Fix the ``rst_prolog = f" .. |python_requires| replace:: {setup_cfg['options']['python_requires']}"``, which is currently commented out (it otherwise adds a blank line at the top of every page) 
- get the badges to float right on the index page (I think the class's use of ``::before`` makes the float not work)
- ``docs/_static/theme_overrides.css`` can definitely be shortened, I just duplicated for now
- clean up a bit from my testing and old stuff 

Notice that the theme now supports dark mode. A future PR can set a different color scheme more suitable for dark mode.
image

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -28,55 +28,55 @@ Packages
 ========
 
 The Ocean SDK includes the :ref:`dwave_cli` and the following packages:
-
+   
 .. packages-start-marker
 
 .. dropdown::  **dimod** --- Quadratic models: BQM, DQM. \
-   :link-badge:`https://docs.ocean.dwavesys.com/en/stable/docs_dimod/sdk_index.html, "docs", cls=badge-primary fr text-white` \
-   :link-badge:`https://github.com/dwavesystems/dimod, "code", cls=badge-secondary fr text-white`
-
+   :bdg-link-primary:`docs <https://docs.ocean.dwavesys.com/en/stable/docs_dimod/sdk_index.html>` \
+   :bdg-link-secondary:`code <https://github.com/dwavesystems/dimod>` 
+ 
    Shared API for binary quadratic :term:`sampler`\ s. Provides a binary quadratic model (BQM) class
    that contains :term:`Ising` and quadratic unconstrained binary optimization (:term:`QUBO`) models
    used by samplers such as the D-Wave system. Also provides utilities for constructing new samplers
    and composed samplers.
 
 .. dropdown:: **dwavebinarycsp** --- Generates BQMs from constraint satisfaction problems. \
-   :link-badge:`https://docs.ocean.dwavesys.com/en/stable/docs_binarycsp/sdk_index.html,"docs", cls=badge-primary fr text-white` \
-   :link-badge:`https://github.com/dwavesystems/dwavebinarycsp,"code", cls=badge-secondary fr text-white`
+   :bdg-link-primary:`docs <https://docs.ocean.dwavesys.com/en/stable/docs_binarycsp/sdk_index.html>` \
+   :bdg-link-secondary:`code <https://github.com/dwavesystems/dwavebinarycsp>`
 
    Library to construct a binary quadratic model from a constraint satisfaction
    problem with small constraints over binary variables.
 
 .. dropdown:: **dwave-cloud-client** --- API client to D-Wave solvers. \
-   :link-badge:`https://docs.ocean.dwavesys.com/en/stable/docs_cloud/sdk_index.html, "docs", cls=badge-primary fr text-white` \
-   :link-badge:`https://github.com/dwavesystems/dwave-cloud-client, "code", cls=badge-secondary fr text-white`
+   :bdg-link-primary:`docs <https://docs.ocean.dwavesys.com/en/stable/docs_cloud/sdk_index.html>` \
+   :bdg-link-secondary:`code <https://github.com/dwavesystems/dwave-cloud-client>`
 
    Minimal implementation of the REST interface used to communicate with D-Wave
    :term:`Sampler` API (SAPI) servers.
 
 .. dropdown:: **dwave-gate** --- Package for quantum circuits. \
-   :link-badge:`https://docs.ocean.dwavesys.com/en/stable/docs_gate/sdk_index.html, "docs", cls=badge-primary fr text-white` \
-   :link-badge:`https://github.com/dwavesystems/dwave-gate, "code", cls=badge-secondary fr text-white`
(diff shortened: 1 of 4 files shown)


JoelPasvolsky:
Looks like I might need to also either update CircleCI to use a new Python version to run doc builds or use an earlier version of Sphinx

arcondello:
It's configured here: https://github.com/dwavesystems/dwave-ocean-sdk/blob/180ec75f6fcda0d7515f8977d8b07eb9e1406afb/.circleci/config.yml#L134

What does RTD currently use?

arcondello:
Looks like it used 3.8 by default, but [it is configurable](https://blog.readthedocs.com/migrate-configuration-v2/).

JoelPasvolsky:
> Looks like it used 3.8 by default, but [it is configurable](https://blog.readthedocs.com/migrate-configuration-v2/).

Yes, I actually recently [made a PR](https://github.com/dwavesystems/dwave-ocean-sdk/pull/272/files) to move to this configuration

arcondello:
Then yes! We can use whatever version you need :laughing:

JoelPasvolsky:
The generated docset can be seen [here](https://output.circle-artifacts.com/output/job/57a31f38-bdca-41b7-85d2-03bd4b715c93/artifacts/0/docs/_build/html/index.html)

arcondello:
So most of this is bike shedding but...

For comparison two images (found [here](https://output.circle-artifacts.com/output/job/2e25deaf-989c-4c4f-a82e-6373abec17a8/artifacts/0/docs/_build/html/overview/cpu.html))

### New Docs

image

### Old Docs

image

### Comments

1) I don't love that the links are underlined. E.g.
image
image
2) I prefer the old pink inline code to the orange
3) I don't love the new python syntax coloring
4) The code link is using green rather than blue E.g.
image
image
5) This isn't int he example image, but the bar at the top has weird coloring when I mouse over. E.g. 
image

JoelPasvolsky:
> 1. I don't love that the links are underlined. E.g.

It's simple to remove the link underlining (`a {text-decoration: none;}` as a theme override) but I believe they added the underline to help people with visual impairments recognize hyperlinks so I propose to keep that. 

>  2. I prefer the old pink inline code to the orange

I restored the previous color (`--inline-code-pink: rgb(232, 62, 140);`) based on the original `pydata_sphinx_theme==0.8.0` code. Again, I believe the added backgrounding is to help with visual impairments, so I left that.
image

> 3. I don't love the new python syntax coloring

Apparently they are using [GitHub code block style](https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/theme-elements.html#code-blocks) now. I don't think we want to try restyling that ourselves

> 4. The code link is using green rather than blue E.g.

Ha, I figured out how to fix that to my huge surprise: `--pst-color-inline-code-links: var(--dwave-color-ocean-blue);` 
image

> 5. This isn't int he example image, but the bar at the top has weird coloring when I mouse over. E.g.

I don't know how to change that one :shrug:

arcondello:
All makes sense to me, thanks

arcondello (approved):

JoelPasvolsky:
@randomir & @arcondello, see [this rendering](https://sdk-docs.readthedocs.io/en/latest/) and let me know if I should go ahead and merge [this commit](https://github.com/JoelPasvolsky/dwave-ocean-sdk/pull/1) into my branch.

arcondello:
I personally prefer it, though I would take a 2x3 or 3x3 subset of the packages to highlight.

JoelPasvolsky:
> I personally prefer it, though I would take a 2x3 or 3x3 subset of the packages to highlight.

It's nicely flexible, we could build off something like this: 

image

docs/index.rst
```diff
@@ -125,31 +125,31 @@ The Ocean SDK includes the :ref:`dwave_cli` and the following packages:
    It also includes several useful composites—layers of pre- and post-processing—that
    can be used with DWaveSampler to handle :term:`minor-embedding`, optimize chain strength, etc.
 
-.. dropdown:: **minorminer** --- Minor-embeds graphs. \
-   :link-badge:`https://docs.ocean.dwavesys.com/en/stable/docs_minorminer/source/sdk_index.html, "docs", cls=badge-primary fr text-white` \
-   :link-badge:`https://github.com/dwavesystems/minorminer, "code", cls=badge-secondary fr text-white`
+.. dropdown:: `minorminer <https://docs.ocean.dwavesys.com/en/stable/docs_minorminer/source/sdk_index.html>`_---Minor-embeds graphs.
```
randomir, line 128: Use a ref here?

randomir (approved):
LGTM, apart from a few nits:
- add space between package name, dash and description 
image
- code/docs pills feel unrefined/redundant (especially on hover).
image
 I'd remove them completely, and just add a link to source in a new line, e.g.:
image

JoelPasvolsky:
> I personally prefer it, though I would take a 2x3 or 3x3 subset of the packages to highlight.

Another option to consider is the use of tabs for (default) clean look and (alternative tab) full list of packages:

image

docs/index.rst
```diff
@@ -28,79 +28,79 @@ Packages
...
 
 The Ocean SDK includes the :ref:`dwave_cli` and the following packages:
-
+   
 .. packages-start-marker
 
-.. dropdown::  **dimod** --- Quadratic models: BQM, DQM. \
-   :link-badge:`https://docs.ocean.dwavesys.com/en/stable/docs_dimod/sdk_index.html, "docs", cls=badge-primary fr text-white` \
-   :link-badge:`https://github.com/dwavesystems/dimod, "code", cls=badge-secondary fr text-white`
+.. dropdown::  `dimod <https://docs.ocean.dwavesys.com/en/stable/docs_dimod/sdk_index.html>`_---Quadratic models: BQM, DQM.
```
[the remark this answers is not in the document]
JoelPasvolsky: We need to add an `.. _index_dimod:` line in dimod, it's the one package where that is missing and I updated to CQM in the [front-page PR](https://github.com/JoelPasvolsky/dwave-ocean-sdk/pull/1) already.

docs/index.rst
JoelPasvolsky: We need to add an `.. _index_minorminer:` line in the package

arcondello:
> Another option to consider is the use of tabs for (default) clean look and (alternative tab) full list of packages

I'd even say the first tab could be boxes and the second tab a list.

PR #282, opened 2023-12-14
Update docs build system for python 3.12

randomir:
`pkg_resources` is not available in python 3.12 by default, as setuptools are not installed by default anymore. Furthermore, its usage has been deprecated in 3.12, to be replaced with `importlib.metadata` and `packaging` functionality.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -10,7 +10,9 @@ import os
 import sys
 import subprocess
 import inspect
-import pkg_resources
+
+from importlib.metadata import Distribution
+from packaging.requirements import Requirement
 
 sdk_directory = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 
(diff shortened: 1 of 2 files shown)


docs/conf.py
```diff
@@ -213,9 +215,9 @@ def setup(app):
               'embedding': 'dwave-system',
```
JoelPasvolsky, line 215: For Windows build with Python 3.10,  I'm hitting some problems:

* [setup_cfg['metadata']](https://github.com/dwavesystems/dwave-ocean-sdk/blob/c04a51752dfbe5e6f460752dc51617f15d620d7f/docs/conf.py#L30) gives a key error
* Line 219 `pkgs = [Distribution.from_name(req.name) for req in reqs]` gives `PackageNotFoundError: No package metadata was found for dwave-gate` that I think is because you switched `requires(extras=['all']` to `requires`

randomir: Well, luckily, we're not building our docs on Windows :laughing: 

Can you make sure you ran `pip install .[all]` in your environment, just like we do in the CI job?

Note: `importlib.metadata.Distribution.requires` is now a property, hence it does not accept any parameters, but the code should work if the package is installed.

arcondello (approved):

docs/conf.py
JoelPasvolsky: Actually it turns out that if you run the above code line in the correct environment rather than some other random terminal, it in fact works.

JoelPasvolsky (approved):
Okay, these changes work for me in Win Py 3.10 environment, thank you!

thisac (approved):
LGTM

PR #300, opened 2024-06-21
Clean up non-submodule build warnings

JoelPasvolsky:
Currently building Ocean SDK with `make html` produces close to 150 warnings, a majority of type `WARNING: more than one target found for cross-reference`. This causes two problems:

1. Links across modules are to an arbitrary target of the several similarly named properties (e.g. dimod and dwave-hybrid `SampleSet` or various `nodelist` attributes).  
2. We miss new significant warnings because no one is looking through 150 warnings

We had originally accepted these warnings out of the expectation that Sphinx would strengthen inheritance disambiguation but this [GitHub issue on Sphinx](https://github.com/sphinx-doc/sphinx/issues/10400) is now two years old. 

The updates in this and the following PRs reduce the number of warnings from **~150 to under 20**:  

- ✔️ [dimod PR1](https://github.com/dwavesystems/dimod/pull/1372), [dimod PR2](https://github.com/dwavesystems/dimod/pull/1378), and [dimod PR3](https://github.com/dwavesystems/dimod/pull/1377)
- ✔️ [dwave-system PR](https://github.com/dwavesystems/dwave-system/pull/524)
- ✔️ [dwave-cloud-client PR](https://github.com/dwavesystems/dwave-cloud-client/pull/633)
- ✔️ [dwave-hybrid PR](https://github.com/dwavesystems/dwave-hybrid/pull/294)
- ✔️ [dwave-preprocessing PR1](https://github.com/dwavesystems/dwave-preprocessing/pull/139) and [dwave-preprocessing PR2](https://github.com/dwavesystems/dwave-preprocessing/pull/138)
- ✔️ [dwave-gate PR](https://github.com/dwavesystems/dwave-gate/pull/49)
- ✔️ [dwave-inspector PR](https://github.com/dwavesystems/dwave-inspector/pull/178)
- ✔️ [penaltymodel PR1](https://github.com/dwavesystems/penaltymodel/pull/149) and [penaltymodel PR2](https://github.com/dwavesystems/penaltymodel/pull/150)
- ✔️ [dwave-samplers PR](https://github.com/dwavesystems/dwave-samplers/pull/63)
- ✔️ [dwave-networkx PR](https://github.com/dwavesystems/dwave-networkx/pull/241)


JoelPasvolsky:
I'm proposing here for consideration to build C++ docs in CircleCI to expand what is tested before SDK release. 

Originally I wanted this to also remove the half dozen minorminer warnings due to lack of C++ docs, but building those added ~500 new C++ duplicate references warnings. (I can suppress those in doxygen but not sphinx, there's no `suppress_warnings` type for C++.) 

So [this commit](https://github.com/dwavesystems/dwave-ocean-sdk/pull/300/commits/9886f24bdcb4410d97136e1dc141d16c1d2ea158) builds just dimod and dwave-preprocessing C++. (It is using a temporary hack that can be replaced with updated Doxygen files in those repos.) Minorminer building needs more work outside this PR. 

With the related [PRs above](https://github.com/dwavesystems/dwave-ocean-sdk/pull/300#issue-2367194323), we should be down from 150 warnings to a handful (and I can still reduce those), mainly on minorminer's lack of C++ docs and old Doxygen files (easy to fix). 

However, the C++ produces `dot` errors (about 100). Still those are easily identified and can be ignored when checking the build log.    

I want some feedback on the proposal to build C++ docs in CI before I update Doxyfiles in additional PRs.

.circleci/config.yml
```diff
@@ -160,7 +160,8 @@ jobs:
           name: build the docs
           command: |
             . env/bin/activate
-            make -C docs/ html
+            sudo apt update && sudo apt install doxygen
```
randomir, line 163: To fix `dot` errors:
```suggestion
            sudo apt update && sudo apt install doxygen graphviz
```

docs/conf.py
```diff
@@ -133,11 +133,22 @@
 breathe_default_members = ('members', )
 breathe_default_project = "minorminer"
 
-# we want to build the c++ docs in RTD
-if os.environ.get('READTHEDOCS', False):
+# we want to build the c++ docs in RTD and CircleCI: 
+if os.environ.get('READTHEDOCS', False):
```
randomir, line 137: trailing spaces

randomir (approved):
LGTM.

Generally, I think it's good to test in advance of the release build (d'oh!), but I also think our current Ocean docs build flow could benefit from a refresh. Ideally, we would build as much of Ocean docs (if not all) on every docs change, in any Ocean package -- not only when packages/submodules are updated here.

TL;DR: I agree we go with this for now.

randomir:
We should also reevaluate the benefits of running linkcheck on every docs build vs the cost. It adds 12+ minutes, making doctest job run for over **17 minutes**! That's ridiculous compared to 1 minute runtime of other jobs.

Even more so because we ignore the linkcheck results anyway.

JoelPasvolsky:
> we should run linkcheck job like we run synk/blackduck, nightly.

I wonder why it takes so long to run, that doesn't make sense, let me see if it's doing something weird.
But otherwise agree it can just run periodically

randomir:
> I wonder why it takes so long to run, that doesn't make sense, let me see if it's doing something weird. But otherwise agree it can just run periodically

I assume it takes so long because it has to open each URL to see if link work.. and when they don't, it retries, etc.

Possibly we can improve performance by tweaking request parameters (timeout, number of retries, etc).

randomir:
Btw, test failure is unrelated, and fixed with https://github.com/dwavesystems/dwave-ocean-sdk/pull/301. Feel free to rebase once we merge it.

docs/conf.py
JoelPasvolsky: Fixed, thanks

JoelPasvolsky:
> I assume it takes so long because it has to open each URL to see if link works..

Today's run dropped from 17 minutes to 43 seconds. Let's wait a bit before removing the linkcheck

JoelPasvolsky:
In my local build there remain the following warnings only:
* 8 warnings due to reno duplicates. Not sure we have an easy way to get rid of these. 
* 9 warnings from minorminer. These are expected because we're not currently building minorminer's C++ docs in CircelCI (too many build warnings)
* a couple of warnings for duplicates between dwave-samplers and dwave-neal & dwave-greedy, which will go away 

I think this is reasonable for this PR.

randomir:
Sounds good, @JoelPasvolsky. Would you consider this PR ready for merging then?

PR #302, opened 2024-07-25
Update packages for 7.1.0

thisac:
Update packages and subprojects to latest releases.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,20 +26,20 @@ description = Software development kit for open source D-Wave tools
 
 [options]
 install_requires =
-    dimod==0.12.15
-    dwave-cloud-client==0.12.0
+    dimod==0.12.16
+    dwave-cloud-client==0.12.1
     dwave-greedy==0.3.0
     dwave-hybrid==0.6.11
     dwave-inspector==0.5.0.post0
     dwave-neal==0.6.0
     dwave-networkx==0.8.15
-    dwave-optimization==0.1.0
-    dwave-preprocessing==0.6.5
-    dwave-samplers==1.2.0
+    dwave-optimization==0.2.0
+    dwave-preprocessing==0.6.6
+    dwave-samplers==1.3.0
     dwave-system==1.25.0
     dwave-tabu==0.5.0
     dwavebinarycsp==0.3.0
-    minorminer==0.2.14
+    minorminer==0.2.15
     penaltymodel==1.1.0
     # exclude pyqubo until it supports py312
     # pyqubo==1.4.0
@@ -48,4 +48,4 @@ python_requires = >=3.8
 
 [options.extras_require]
 all =
-    dwave-gate==0.3.1
+    dwave-gate==0.3.2
--- a/setup.cfg
+++ b/setup.cfg
@@ -28,6 +28,7 @@ description = Software development kit for open source D-Wave tools
 install_requires =
     dimod==0.12.16
     dwave-cloud-client==0.12.1
+    dwave-gate==0.3.2
     dwave-greedy==0.3.0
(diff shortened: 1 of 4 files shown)


setup.cfg
```diff
@@ -48,4 +48,4 @@ python_requires = >=3.8
 
 [options.extras_require]
 all =
-    dwave-gate==0.3.1
+    dwave-gate==0.3.2
```
arcondello, line 51: Did we want to promote `dwave-gate` at the same time?

arcondello (approved):

setup.cfg
thisac: We could, should we? Unless anyone thinks otherwise.

arcondello: :shrug: No preference either way.

arcondello:
Doctest failures look like they are mostly from NumPy object printing changes in NumPy 2.0. Maybe makes sense to bound `Numpy<2.0` in the doctests for now?

setup.cfg
thisac: Alright, let's promote it now then. I'll think about how to best structure other extras (specifically the QIR dependency in dwave-gate) later.

arcondello:
Oh, also, I think we should add 
```python
autodoc_type_aliases = {
    'numpy.typing.ArrayLike': 'numpy.typing.ArrayLike',
}
```
 to `conf.py` for slightly more readable docs in some places. There are other places we could/should add aliases as well, but that one will definitely help with `dwave-optimization` generator docs.

setup.cfg
```diff
@@ -45,7 +46,3 @@ install_requires =
     # pyqubo==1.4.0
 packages = dwaveoceansdk
 python_requires = >=3.8
-
-[options.extras_require]
-all =
```
arcondello, line 50: I would leave `all` in with nothing. Because otherwise any `pip install dwave-ocean-sdk[all]` out there will break.

thisac: Ah, good point.

setup.cfg
```diff
@@ -26,26 +26,23 @@ description = Software development kit for open source D-Wave tools
...
     dwave-inspector==0.5.0.post0
     dwave-neal==0.6.0
     dwave-networkx==0.8.15
-    dwave-optimization==0.1.0
-    dwave-preprocessing==0.6.5
-    dwave-samplers==1.2.0
+    dwave-optimization==0.2.0
+    dwave-preprocessing==0.6.6
+    dwave-samplers==1.3.0
     dwave-system==1.25.0
```
randomir, line 40: We could also release new dwave-system, but since there're only minor test and docs improvements, we can skip it this time as well.

randomir (approved):
lgtm

setup.cfg
thisac: Seems like only a warning is produced when calling an extra that doesn't exist though, so probably no big harm in not having it. It's there as an empty extra now anyway.

thisac:
@arcondello I don't see what `autodoc_type_aliases = {'numpy.typing.ArrayLike': 'numpy.typing.ArrayLike'}` would do. Shouldn't the key be a _different_ alias than the value?

arcondello:
Docs on this branch
image

Docs with the alias
image


This is basically saying "replace any instance of `numpy.typing.ArrayLike` with the literal string `numpy.typing ArrayLike`. Which prevents it from (unhelpfully) expanding it out.

setup.cfg
arcondello: Ah, it used to be an error, but looks like they changed it to a warning in [63.4.2](https://setuptools.pypa.io/en/stable/history.html#v63-4-2).

PR #307, opened 2024-09-11
Update packages for 8.0.0

randomir:


thisac (approved):
Do you want to bump the version here as well or do that separately?

randomir:
I prefer to bump the version in a separate commit.

Issue #192, opened 2022-02-25
Deprecation guideline


arcondello:
It would be useful for the docs to have our deprecation policy spelled out.

Also having an official "deprecation text" would be useful so we can be more consistent across Ocean. I was speaking with @JoelPasvolsky and we recommend something like 
```python
warnings.warn(
    "<feature> is deprecated since X.Y.Z and will be removed in <package> (X+2).0.0. "
    "Use <suggested alternative>."
    DeprecationWarning, stacklevel=...
)
```
in the code and
```rst
.. deprecated:: X.Y.Z

    <feature> will be removed in <package> (X+2).0.0. Use <suggested alternative>.
```
in the docstring. `.. deprecated` prepends `Deprecated since version X.Y.Z:` to the notice.

PR #312, opened 2024-10-03
Remove obsoleted classical-sampler packages from docs

JoelPasvolsky:
``dwave-greedy``, ``dwave-tabu``, and ``dwave-neal`` were deprecated in ``dwave-ocean-sdk`` version 6.1.0 in favor of [dwave-samplers](https://docs.ocean.dwavesys.com/en/stable/docs_samplers/index.html#index-dwave-samplers) and was slated to be removed in ``dwave-ocean-sdk`` 8.0.0.


arcondello:
Also need to be removed from https://github.com/dwavesystems/dwave-ocean-sdk/blob/master/setup.cfg

docs/packages.rst
```diff
@@ -171,17 +171,14 @@ Packages
     docs_dimod/sdk_index
     docs_binarycsp/sdk_index
     docs_cloud/sdk_index
-    docs_gate/sdk_index
```
randomir, line 174: I believe you meant to remove the next line, not this one!

JoelPasvolsky: oops. Thanks!

randomir:
> Also need to be removed from https://github.com/dwavesystems/dwave-ocean-sdk/blob/master/setup.cfg

That will remove `greedy`, `neal` and `tabu` top-level packages from Ocean as well. Are we sure we want to do that? (now?)

arcondello:
I don't see why not. They've been deprecated for a long time.

JoelPasvolsky:
> remove `greedy`, `neal` and `tabu` top-level namespaces

Based on the doctest's passing, likely the Ocean examples are all updated to use the `dwave-sampler` package, not sure about [dwave-examples](https://github.com/dwave-examples) code though

arcondello:
Quick github search doesn't turn up any.

arcondello:
https://github.com/dwave-examples/frequency-selection as well.

arcondello (approved):
Regardless, I think we should proceed. We said we'd deprecate as part of 8.x, and here we are. So I am OK with moving forward.

randomir:
`neal` is used in [7 examples](https://github.com/search?q=org%3Adwave-examples+%22from+neal+import%22&type=code), `tabu` in [2 examples](https://github.com/search?q=org%3Adwave-examples+%22from+tabu+import%22&type=code), and `greedy` in none.

But across GitHub, only [`neal`](https://github.com/search?q=%22from+neal+import%22+&type=code) and [`tabu`](https://github.com/search?q=%22from+tabu+import%22+&type=code) are used in **few hundred files**. Searching for `greedy` usage is harder because it's not as unique.

We should consider introducing `neal`/`tabu`/`greedy` namespaces to `dwave-samplers`, not just for backwards-compatibility, but also for convenience.

arcondello:
I think we'd just be prolonging the issue. A quick look shows many of those files haven't been updated in years. Fewer namespaces, fewer installs, etc are all good in my books. If we really want all demos updated before deploying this change in the SDK then sure, we can do that. Though the fact that most of them have broken CI seems much more pressing to me.

If we weren't comfortable removing them in 0.8.0, then we shouldn't have said we would.

randomir:
I absolutely agree, we should drop it as planned. But the issue is this deprecation was pretty silent -- I can't find it in docs, and there's no `DeprecationWarning` on import from these deprecated namespaces. So, most likely it will come as a big surprise to users at this point. :shrug:

JoelPasvolsky:
> can't find it in docs

https://docs.ocean.dwavesys.com/en/stable/docs_greedy/sdk_index.html
https://docs.ocean.dwavesys.com/en/stable/docs_neal/sdk_index.html
https://docs.ocean.dwavesys.com/en/stable/docs_tabu/sdk_index.html

arcondello:
We definitely should be better about our deprecations going forward. A warning on import would have been better.

Even so, I am quite comfortable with the risk. And if we get a bug report of three we can always restore them. But for now I am strongly inclined to bias towards slimming things down. And this seems like a very safe place to start.

randomir:
Yes, I found it now, when opening packages from [here](https://docs.ocean.dwavesys.com/en/stable/packages.html), but they're not present in the list of packages [on the main page](https://docs.ocean.dwavesys.com/en/stable/) (that I usually use).

JoelPasvolsky:
> not present in the list of packages

Right, they were removed there rather than noticed as deprecated

thisac:
> `neal` is used in [7 examples](https://github.com/search?q=org%3Adwave-examples+%22from+neal+import%22&type=code), `tabu` in [2 examples](https://github.com/search?q=org%3Adwave-examples+%22from+tabu+import%22&type=code), and `greedy` in none.
> 
> But across GitHub, only [`neal`](https://github.com/search?q=%22from+neal+import%22+&type=code) and [`tabu`](https://github.com/search?q=%22from+tabu+import%22+&type=code) are used in a **few hundred files**, according to GH search at least. Searching for `greedy` usage is harder because it's not as unique.
> 
> We should consider introducing `neal`/`tabu`/`greedy` namespaces to `dwave-samplers`, not just for backwards-compatibility, but also for convenience.

I'll make sure to have a look at updating (or archiving) these demos. A couple of these are still being used.

randomir:
Ok, let's do it then.

But let's also keep a reference / note these top-level namespaces (and packages) were removed in Ocean 8? (or rather 9 since we forgot to remove them in 8). Just so that googling "neal" points users to a note they need to upgrade..

arcondello:
> On that note, we should be using .. versionadded::, .. versionchanged:: and [.. versionremoved::](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionremoved) more. I find it extremely useful when reading other docs (as rare as that might be 😄).

Agree I like them.

Though I long for a nice way to leave the version unspecified until the version is actually tagged. I hate having to guess whether the next version will be a major or minor or patch version when writing the docs.

I've thought about adding them with `TODO` or some other searchable string and then updating the version when updating `__version__` but that seems like a pain. :shrug:

randomir:
Yes, projecting version number at docs write time is hard and unreliable, especially for smaller things we don't have a master plan for. But what I find equally frustrating (and users probably way more) -- is determining Ocean version from package version, and deciding should I quote package version or Ocean version in the docs. One benefit of a monorepo, for sure :grin:.

arcondello:
> determining Ocean version from package version, and deciding should I quote package version or Ocean version in the docs

We could mention the package name in the note (suggested in https://github.com/dwavesystems/dwave-ocean-sdk/issues/192). A bit redundant but nicely explicit. We could also mess about with the sphinx build css to insert it when building. Probably not too hard to do :thinking:

PR #314, opened 2024-11-09
Add pyqubo 1.5.0 to support Python 3.12

kotarotanahashi:
Since PyQUBO now supports Python 3.12 and 3.13, we would appreciate it if you could add it back to the D-Wave Ocean SDK. Thank you for your consideration. 

Here is the latest release of PyQUBO:
https://pypi.org/project/pyqubo/


arcondello:
Hi @kotarotanahashi. Thank you for your PR. Unfortunately we must decline. We'd like to limit the SDK as much as possible to D-Wave packages and their direct dependencies.

PR #316, opened 2024-12-06
Drop AppVeyor in favor of CircleCI for Windows tests

randomir:


randomir:
We can unify python versions after `orb-ocean` is adapted to resolve the latest patch version from minor version.

arcondello (approved):

PR #339, opened 2025-04-25
Fix footnote on fast-anneal resolution

JoelPasvolsky:
A previous PR (for DOC-742) updated the footnote [here](https://docs.dwavequantum.com/en/latest/quantum_research/solver_properties_all.html#fast-anneal-time-range) but missed [this one](https://docs.dwavequantum.com/en/latest/quantum_research/solver_parameters.html#anneal-schedule). Also, "specify times to a maximum resolution of 5%" in the second link is a typo.
Additionally, the `start_time_granularity` content is duplicated while the two footnotes are not single sourced.
I'll add [Emile ](https://github.com/ehoskinson) as a reviewer once he's set up.


qwriter (approved):
LGTM!

ehoskinson (approved):
Looks good!

PR #324, opened 2025-03-07
Unify system and Ocean content

JoelPasvolsky:
This update is dependent on the following PRs to submodules (all now merged):

* [dimod](https://github.com/dwavesystems/dimod/pull/1400) + [dimod dot](https://github.com/dwavesystems/dimod/pull/1401)
* [dwave-cloud-client](https://github.com/dwavesystems/dwave-cloud-client/pull/687)
* [dwave-preprocessing](https://github.com/dwavesystems/dwave-preprocessing/pull/146) + [filenames renaming](https://github.com/dwavesystems/dwave-preprocessing/pull/148) + [dwave-preprocessing dot](https://github.com/dwavesystems/dwave-preprocessing/pull/149)
* [dwavebinarycsp](https://github.com/dwavesystems/dwavebinarycsp/pull/125) +[filenames renaming](https://github.com/dwavesystems/dwavebinarycsp/pull/126)
* [dwave-inspector](https://github.com/dwavesystems/dwave-inspector/pull/185)
* [dwave-optimization](https://github.com/dwavesystems/dwave-optimization/pull/261)
* [dwave-hybrid](https://github.com/dwavesystems/dwave-hybrid/pull/305)
* [dwave-samplers](https://github.com/dwavesystems/dwave-samplers/pull/76)
* [penaltymodel](https://github.com/dwavesystems/penaltymodel/pull/154)
* [dwave-gate](https://github.com/dwavesystems/dwave-gate/pull/55)
* [dwave-networkx](https://github.com/dwavesystems/dwave-networkx/pull/250)
* [minorminer](https://github.com/dwavesystems/minorminer/pull/262)
* [dwave-system](https://github.com/dwavesystems/dwave-system/pull/560)

PR #342, opened 2025-05-02
Update Leap authorization

JoelPasvolsky:
Closes DOC-800, DOC-467, and DOC-926 to updated the [Leap Authorization](https://docs.dwavequantum.com/en/latest/ocean/leap_authorization.html) page to use `dwave setup --auth` instead of `dwave auth login`, to include steps for accounts with multiple projects, and to reference the [Ocean Installation](https://docs.dwavequantum.com/en/latest/ocean/install.html) page as a prerequisite.


docs/ocean/leap_authorization.rst
```diff
@@ -32,18 +33,78 @@ commands and all their options with the CLI's ``--help`` option.
     using the secure `OAuth 2.0 <https://oauth.net/2/>`_ code exchange, by
     running one of the following commands from your terminal.
```
qwriter, line 34: Tabs are an excellent feature! Perhaps a phrase introducing the tabs would help clarify their titles? For example, "... terminal, depending on whether you have only one or multiple Leap projects:"?

docs/ocean/leap_authorization.rst
```diff
@@ -32,18 +33,78 @@ commands and all their options with the CLI's ``--help`` option.
...
-            dwave auth login
+        .. tab-item:: Single Leap Project
+            :name: tab_authorize_single_leap_project
+            :selected:
 
-    -   In an environment such as a **cloud IDE**, where access to
-        `localhost <https://en.wikipedia.org/wiki/Localhost>`_ addresses from
-        your browser might be blocked, you can initiate the alternative OAuth
-        *out-of-band* flow::
+            If your account has more than one project, the
```
qwriter, line 42: With the suggested introductory phrase, perhaps this paragraph won't be needed.

docs/ocean/leap_authorization.rst
```diff
@@ -23,7 +23,8 @@ environment.
 
 This procedure uses the :ref:`Ocean software CLI <ocean_dwave_cli>` commands
 that you enter into your terminal. You can see help documentation for these
-commands and all their options with the CLI's ``--help`` option.
+commands and all their options with the CLI's ``--help`` option. This CLI is
+installed as part of the :ref:`Ocean SDK installation <ocean_install>`.
```
qwriter, line 27: Or docker image?

qwriter (approved):
A few suggestions and one typo; otherwise, LGTM.

docs/ocean/leap_authorization.rst
JoelPasvolsky: I went back and forth on this, and in the end not adding more to the wall of text that precedes the procedure won out. My thinking is that most users will be single-project so no harm if they miss/ignore the 2nd tab, those looking will find it easily enough, and, as we discussed, keeping it sparse is high priority.  
Still, I'll let it rest over the weekend and then see if I can reduce/hide/redistribute those introductory 169 words, opening space for an explanatory sentence to the tab names.

docs/ocean/leap_authorization.rst
JoelPasvolsky: Thanks, removed

docs/ocean/leap_authorization.rst
```diff
@@ -32,18 +33,78 @@ commands and all their options with the CLI's ``--help`` option.
...
+
+            .. doctest::
+                :skipif: True
+
+                    dwave setup --oob --project PROJ3
+
+            You can always switch your environment to use the API token of
+            another project later. For example, if you ran the procedure of the
+            :ref:`tab_authorize_single_leap_project` tab with the ``PROJ2``
+            project set to active in Leap , you can at any time switch to the
```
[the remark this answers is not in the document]
JoelPasvolsky: Forgot this, fixed now, thanks

docs/ocean/leap_authorization.rst
```diff
@@ -10,40 +10,92 @@ and higher enable Leap\ |tm| authentication using the secure
...
-This procedure uses the :ref:`Ocean software CLI <ocean_dwave_cli>` commands
-that you enter into your terminal. You can see help documentation for these
-commands and all their options with the CLI's ``--help`` option.
+The following procedure authorizes the Ocean software access to your account in
+the Leap service and enables storing your SAPI token in your development
+environment. Ocean software authorization to access the Leap service persists
+across subsequent development sessions for this development environment.
+
+This procedure uses the :ref:`Ocean software CLI <ocean_dwave_cli>`
+commands\ [#]_ that you enter into your terminal.
```
randomir, line 24: Broken ref. It shows `[1]` in the [built docs](https://output.circle-artifacts.com/output/job/6efee9a3-37d6-41d1-a4f8-c93a5630cf49/artifacts/0/docs/_build/html/ocean/leap_authorization.html), but it's linked with a hidden footnote on the second tab below. Which is also irrelevant for this "CLI commands" ref.

docs/ocean/leap_authorization.rst
```diff
@@ -10,40 +10,92 @@ and higher enable Leap\ |tm| authentication using the secure
...
 
 #.  For any new developer environment (for example, a Python
     `virtual environment <https://docs.python.org/3/library/venv.html>`_
     or a `GitHub Codespaces <https://docs.github.com/codespaces>`_ *codespace*),
-    authorize the Ocean software access to your account in the Leap service,
-    using the secure `OAuth 2.0 <https://oauth.net/2/>`_ code exchange, by
-    running one of the following commands from your terminal.
+    run one of the following commands from your terminal. If your account has
+    more than one project, the :ref:`tab_authorize_multiple_leap_projects` tab
```
randomir, line 30: This creates a link to table, and jumps to the tab bar, but it doesn't actually open the second tab.

docs/ocean/leap_authorization.rst
```diff
@@ -10,40 +10,92 @@ and higher enable Leap\ |tm| authentication using the secure
...
+                :skipif: True
+
+                    dwave setup --oob --project PROJ3
+
+            You can always switch your environment to use the API token of
+            another project later. For example, if you ran the procedure of the
+            :ref:`tab_authorize_single_leap_project` tab with the ``PROJ2``
+            project set to active in Leap, you can at any time switch to the
+            ``PROJ3`` project with the
+            ``dwave config create --auto --project PROJ3`` command.
```
randomir, line 93: I would keep `--auto-token` instead `--auto`. It's clearer what part of config create is "auto".

docs/ocean/leap_authorization.rst
```diff
@@ -81,7 +133,7 @@ commands and all their options with the CLI's ``--help`` option.
     terminal's "``Authorization code:``" prompt, similar to the representative
     shell lines shown below::
 
-        $ dwave auth login --oob
+        $ dwave setup --oob
```
randomir, line 136: If we decide to go with `setup` instead of `auth login`, then the output below has to be updated accordingly as well.

randomir (changes requested):
`dwave setup` does a lot more than just "authorize access to Leap", so it shouldn't completely replace `dwave auth login` that does the actual authorization on the page that talks about Leap authorization. (The setup command also installs the inspector, and creates the config file.)

I see why the use of `dwave setup` is convenient, but the Leap auth page should explain Leap auth. And we have (two!) "Configuring Access to the Leap Service" page(s) where we can explain how `dwave auth login` followed by `dwave config create --auto-token` can be replaced with `dwave setup --auth`, as part of Ocean setup.

IMO, we should restructure (distill) these three "getting started" pages ("Authorizing Access to the Leap Service" and two "Configuring Access to the Leap Service" pages). They contain way too much info for someone just getting started (e.g. a ref to a low-level cloud-client function `get_configfile_paths()`). I'd move the bulk of this to the CLI (ref) docs, and keep the "getting started" to a single page (with links to more details).

docs/ocean/leap_authorization.rst
JoelPasvolsky: Updated

docs/ocean/leap_authorization.rst
JoelPasvolsky: Updated

docs/ocean/leap_authorization.rst
JoelPasvolsky: Yes, that's a sphinx limitation, perhaps future versions will open the referenced tab

docs/ocean/leap_authorization.rst
JoelPasvolsky: Fixed, it was to the footnote at the bottom but was hijacked by the new footnote ref. Removed the footnote.

JoelPasvolsky:
> `dwave setup` does a lot more than just "authorize access to Leap", so it shouldn't completely replace `dwave auth login`

The main purpose and audience for this page under Getting Started are users looking to set up their environment, it's not meant to be reference documentation for authorization. The purpose of this PR was not streamline this and one part of that was reducing the explanatory preface from 160 words to 100 words, to help funnel impatient users to a procedure that would get them started with as little fuss as possible.

To meet your concerns, which are certainly valid, I've added [this content](https://github.com/dwavesystems/dwave-ocean-sdk/pull/342/commits/fc8c21df5302f6e079c31f739bf8af3f188418fe). That brings the explanatory preface back up to risky 120 words, so please let us stop there.

For your larger recommendations, let's keep the scope of this PR to improving this content and then look at reworking the three GS pages and reference documentation for authorization, if that's needed, separately.

docs/ocean/leap_authorization.rst
```diff
@@ -10,40 +10,94 @@ and higher enable Leap\ |tm| authentication using the secure
...
 
-    -   In an environment such as a **cloud IDE**, where access to
-        `localhost <https://en.wikipedia.org/wiki/Localhost>`_ addresses from
-        your browser might be blocked, you can initiate the alternative OAuth
-        *out-of-band* flow::
+            .. doctest::
+                :skipif: True
 
-            dwave auth login --oob
+                    dwave setup --oob --project PROJ3
```
randomir, line 88: Extra spaces:

image


```suggestion
                dwave setup --oob --project PROJ3
```

randomir:
> The main purpose and audience for this page under Getting Started are users looking to set up their environment, it's not meant to be reference documentation for authorization. The purpose of this PR was to streamline this and one part of that was reducing the explanatory preface from 160 words to 100 words, to help funnel impatient users to a procedure that would get them started with as little fuss as possible.

Sure, if you insist on using `dwave setup` on the "Authorizing Access to the Leap Service" page, then you should also update:
- "Set Up Your Environment" section on "Installing the Ocean SDK" - it has 3 steps, and all three are now done during the "authorizing" step.
- CLI ref / Leap section - this is a "CLI reference" page, yet it contains a lot less details then the getting started page
- CLI ref / Setup section - already uses `--auth` (it probably shouldn't). Also the example output should be refreshed.

JoelPasvolsky:
> then you should also update

💯% agree. These page can be updated to integrate better, that was done only roughly in the [unification PR](https://github.com/dwavesystems/dwave-ocean-sdk/pull/324). That is what I am suggesting [here](https://github.com/dwavesystems/dwave-ocean-sdk/pull/342#issuecomment-2877925385): 

> For your larger recommendations, let's keep the scope of this PR to improving this content and then look at reworking the three GS pages and reference documentation for authorization, if that's needed, separately.

The need is definitely there (the CLI page has outputs from version 6.7.0), just the time is not quite now 😁

randomir:
OK, we can do that in sync with cloud-client docs updates (where CLI docs reside anyway).

randomir (approved):

PR #350, opened 2025-05-23
Fix invalid doctest directive

JoelPasvolsky:
To close [this PR comment](https://github.com/dwavesystems/dwave-ocean-sdk/pull/349#pullrequestreview-2864688741).
From what I [see here](https://docs.python.org/3/library/doctest.html#option-flags), doctest doesn't support a ``IGNORE_RESULT`` directive, and sphinx enables three directives [by default](https://github.com/sphinx-doc/sphinx/blob/1a69059295e297f987b58918bd40a6c93771f1d8/sphinx/ext/doctest.py#L653) but does not add such a thing itself through the `doctest.register_optionflag()`. We could consider adding this ourselves but it is only used here so until such a time, I think we can just show the output for this example.

### DIFF
--- a/docs/industrial_optimization/solver_dqm_parameters.rst
+++ b/docs/industrial_optimization/solver_dqm_parameters.rst
@@ -44,8 +44,10 @@ rock-paper-scissors---to Ocean software's
 >>> win = {"rock": "scissors", "paper": "rock", "scissors": "paper"}
 ...
 >>> dqm = dimod.DiscreteQuadraticModel()
->>> dqm.add_variable(3, label='my_hand')    # doctest: +IGNORE_RESULT
->>> dqm.add_variable(3, label='their_hand') # doctest: +IGNORE_RESULT
+>>> dqm.add_variable(3, label='my_hand')
+'my_hand'
+>>> dqm.add_variable(3, label='their_hand')
+'their_hand'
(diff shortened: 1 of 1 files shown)


randomir:
> From what I [see here](https://docs.python.org/3/library/doctest.html#option-flags), doctest doesn't support a `IGNORE_RESULT` directive, and sphinx enables three directives [by default](https://github.com/sphinx-doc/sphinx/blob/1a69059295e297f987b58918bd40a6c93771f1d8/sphinx/ext/doctest.py#L653) but does not add such a thing itself through the `doctest.register_optionflag()`. We could consider adding this ourselves but it is only used here so until such a time, I think we can just show the output for this example.

Actually, if it's simple to enable the support for this directive, I'd do that instead. 

These extra two lines of output not only are not adding any value, they're reducing it.

JoelPasvolsky:
> if it's simple to enable

The directive does not exist. One would need to create (code) such a thing. I encourage you to make a PR to sphinx adding it there, other users will also find it useful.

randomir:
> > if it's simple to enable
> 
> The directive does not exist. One would need to create (code) such a thing. I encourage you to make a PR to sphinx adding it there, other users will also find it useful.

Oh, I misunderstood you then.

randomir (approved):

PR #356, opened 2025-06-12
Fix readme syntax and add CI build test

randomir:


.circleci/config.yml
```diff
@@ -194,6 +194,30 @@ jobs:
...
+      - checkout
+
+      - run: *create-virtualenv
+
+      - run: &build-package
+          name: build sdist and bdist
+          command: |
+            . env/bin/activate
+            pip install -U pip setuptools wheel
+            python setup.py sdist
```
arcondello, line 211: IMO we should consider using [build](https://github.com/pypa/build). I think it's a bit more modern/general. And tests that the `pyproject.toml` has the right `setuptools` version etc etc.

randomir: Sure, this is the exact same code we use for deploy, just extracted for this test. We can look into modernizing build, and apply it across all packages.

arcondello: Hmm, though I guess that's a bigger change. And we should be consistent with our deploy job. Perhaps a good thing to do in a followup PR

arcondello: hah, jinx!

thisac (approved):
LGTM

.circleci/config.yml
randomir: Done in https://github.com/dwavesystems/dwave-ocean-sdk/pull/356/commits/61b2ec1e85ffc2de1bbab75961d68d5a2daf9e17.

PR #365, opened 2025-06-20
Add qap vignette

eordog:


docs/industrial_optimization/vignette_qap.rst
```diff
@@ -0,0 +1,163 @@
...
+Results
+=======
+
+All problems were run with a time limit of 150 seconds. Results are reported as
+optimality gaps ((energy/best known solution) - 1) when feasible. Infeasible
+solutions correspond to infinite gaps in the median, and if the median is
+infeasible, the data point is not shown in the plot. In order to impose time
+limits on COIN-OR's solver, presolve techniques are turned off. Solvers that
+terminate faster than the time limit are rerun for the full allocated time
+and the minimum objective value found is saved.
```
JoelPasvolsky, line 143:
```suggestion
All problems were run with a time limit of 150 seconds. Results are reported as
optimality gaps :math:`( \frac{\rm energy}{\rm best\: known\: solution} - 1)`
when feasible. Infeasible solutions correspond to infinite gaps in the median,
and if the median is infeasible, the data point is not shown in the plot. In
order to impose time limits on COIN-OR's solver, presolve techniques are turned
off. Solvers that terminate faster than the time limit are rerun for the full
allocated time and the minimum objective value found is saved.
```



What do you think about using :math:`( \frac{\rm energy}{\rm best\: known\: solution} - 1)`
to switch from 
image

to this:

image

A bit cumbersome but looks better to me. Up to you.

JoelPasvolsky (commented):
LGTM, thank you, @eordog!

JoelPasvolsky (approved):
Thank you!

PR #369, opened 2025-07-12
Document changes to deprecate id SAPI solver resource field and replace with identity

qwriter:
Still need to test examples with identity solver resource field.

### DIFF
--- a/docs/leap_sapi/sapi_rest.rst
+++ b/docs/leap_sapi/sapi_rest.rst
@@ -698,11 +698,10 @@ query.\ [#]_
                         identifier is represented by the following JSON
                         structure::
 
-                        "solver": {"name": solver_name, "version": {"graph_id": graph_id}}
+                        {"name": solver_name, "version": {"graph_id": graph_id}}
 
                         where the fields are as follows:
 
-                        *    ``solver``: Parent for solver-related fields.
                         *    ``name``: Name of the solver.
                         *    ``version``: Parent for QPU-version-related fields.
                              This field does not apply to hybrid solvers.
@@ -1685,7 +1684,7 @@ to get a subset of solver fields.
             Status: ONLINE    Load: 0.09
         Advantage_system6.4
             Status: ONLINE    Load: 0.02
-        Advantage2_system1.4
+        Advantage2_system1.5
             Status: ONLINE    Load: 0.58
 
     .. include:: ../shared/sapi_rest.rst
--- a/docs/leap_sapi/sapi_rest.rst
+++ b/docs/leap_sapi/sapi_rest.rst
@@ -1991,11 +1991,11 @@ You should get a valid problem identifier in the response.
 .. doctest:: rest_live
     :skipif: test_api_token_set == False
 
-    >>> graph_id = r2['identity']['version']['graph_id']
+    >>> solver_id = r2['identity']
     >>> session.headers['Accept'] = 'application/vnd.dwave.sapi.problems+json; version=3'
     ...
     >>> r3 = session.post(f"{SAPI_HOME}/problems",
-    ...                   json=[{"solver": {"name": qpu_solver, "version": {"graph_id": graph_id}},
+    ...                   json=[{"solver": solver_id,
     ...                          "label": "QPU REST submission 1",
     ...                          "data": {"format": "qp", "lin": lin, "quad": quad},
     ...                          "type": "ising",
@@ -2139,7 +2139,7 @@ section above.
 .. doctest:: rest_live
     :skipif: test_api_token_set == False or hss_spot_check == False
 
-    >>> hybrid_bqm_solvers =  [r1[i]['identity']['name'] for i in range(len(r1)) if
+    >>> hybrid_bqm_solvers =  [r1[i]['identity'] for i in range(len(r1)) if
     ...    r1[i]['properties']['category'] == "hybrid" and "binary" in r1[i]['identity']['name']]
     >>> bqm_solver = hybrid_bqm_solvers[0]
     >>> print(bqm_solver)                           # doctest: +SKIP
@@ -2163,7 +2163,7 @@ a valid problem identifier in the response.
     >>> session.headers['Accept'] = 'application/vnd.dwave.sapi.problems+json; version=3'
     ...
     >>> r8 = session.post(f"{SAPI_HOME}/problems",
-    ...                   json=[{"solver": {"name": bqm_solver},
+    ...                   json=[{"solver": bqm_solver,
     ...                          "label": "REST Submission to hybrid BQM solver 2",
     ...                          "data": {"format": "ref", "data": problem_data_id},
     ...                          "type": "bqm",
--- a/docs/leap_sapi/sapi_rest.rst
+++ b/docs/leap_sapi/sapi_rest.rst
@@ -1,7 +1,3 @@
-..  |resource_rep_version| replace:: ``<x>`` is the version of the resource
-    representation that the client prefers; for example, ``3``. The Leap service
-    responds with the lastest available version.  
-
 .. _leap_sapi_rest:
 
 ===================
@@ -211,11 +207,11 @@ described on the following tabs.
             :ref:`dimod <index_dimod>` :meth:`~dimod.QuadraticModel.to_file`
             method.
 
-            .. dropdown:: Example serialized file
+        .. dropdown:: Example serialized file
 
-                .. include:: ../shared/sapi_rest.rst
-                    :start-after: start_bqm_file
-                    :end-before: end_bqm_file
+            .. include:: ../shared/sapi_rest.rst
+                :start-after: start_bqm_file
+                :end-before: end_bqm_file
 
     .. tab-item:: cURL
 
@@ -268,9 +264,7 @@ Initiate Upload of Problem Data
 -------------------------------
 
 For quantum-classical hybrid solvers in the Leap service, you must upload your
-problem using the ``$SAPI_HOME/bqm/multipart/*`` endpoints and, optionally,
-the ``application/vnd.dwave.sapi.multipart-initiate+json; version=2.0.0``
-resource representation before submitting a
+problem using the ``$SAPI_HOME/bqm/multipart/*`` endpoints before submitting a
 SAPI request to run the problem. For large problems (starting from several
 megabytes and higher), upload the problem data in multiple parts.
 
@@ -288,6 +282,10 @@ integer---required to store the problem.
 SAPI returns an identifier, of type
 `UUID <https://en.wikipedia.org/wiki/Universally_unique_identifier>`_, for the
 uploaded problem.
+
+You can optionally set the ``Accept`` header to control the
+``application/vnd.dwave.sapi.multipart-initiate+json; version=2.0.0``
+resource representation.
 
 .. tab-set::
 
@@ -357,10 +355,8 @@ Upload Problem Data
 -------------------
 
 To upload a problem, or parts of a large problem, send an HTTP PUT request to
-``bqm/multipart/<problem_data_id>/part/<part>`` (for each part, where ``<part>``
-is an integer beginning at 1) and, optionally, the
-``application/vnd.dwave.sapi.multipart-upload+json; version=2.0.0`` resource
-representation.
+the ``bqm/multipart/<problem_data_id>/part/<part>`` endpoints (for each part,
+where ``<part>`` is an integer beginning at 1).
 
 The POST request body should contain the data for (one part of) the problem,
 encoded as described in the :ref:`dimod <index_dimod>`
@@ -369,6 +365,10 @@ MD5 checksum.
 
 For a multi-part upload, data size for each part should be 5 MB (5242880 bytes),
 except for the last part, which may be smaller.
+
+You can optionally set the ``Accept`` header to control the
+``application/vnd.dwave.sapi.multipart-upload+json; version=2.0.0``
+resource representation.
 
 .. testsetup:: md5
 
@@ -486,13 +486,15 @@ except for the last part, which may be smaller.
 Complete the Problem Upload
 ---------------------------
 
-To complete an upload of problem data, send an HTTP POST request to
-``bqm/multipart/<problem_data_id>/combine`` and, optionally, the
-``application/vnd.dwave.sapi.multipart-combine+json; version=2.0.0``
-resource representation.
+To complete an upload of problem data, send an HTTP POST request to the
+``bqm/multipart/<problem_data_id>/combine`` endpoint.
 
 The POST request body should contain the checksum for the entire problem, as a
 string.
+
+You can optionally set the ``Accept`` header to control the
+``application/vnd.dwave.sapi.multipart-combine+json; version=2.0.0``
+resource representation.
 
 .. dropdown:: Calculating the Checksum
 
@@ -596,13 +598,15 @@ Status of a Problem Upload
 --------------------------
 
 To query the status of an in-progress problem upload, send an HTTP GET request
-to ``bqm/multipart/<problem_data_id>/status`` and, optionally, the
-``application/vnd.dwave.sapi.multipart-status+json; version=2.0.0``
-resource representation.
+to the ``bqm/multipart/<problem_data_id>/status`` endpoint.
 
 The GET request should contain no body.
 
 SAPI returns the status as a string with parts and checksums.
+
+You can optionally set the ``Accept`` header to control the
+``application/vnd.dwave.sapi.multipart-status+json; version=2.0.0``
+resource representation.
 
 .. tab-set::
 
@@ -657,13 +661,15 @@ Submit Problems
 ---------------
 
 To submit problems to a solver in the Leap service, send an HTTP POST request to
-``problems`` and, optionally, the
-``application/vnd.dwave.sapi.problems+json; version=<x>`` resource
-representation.\ [#]_
+the ``problems`` endpoint.
 
 The POST request body should contain the JSON-encoded fields described below.
 When possible, if you have more than one problem, submit them in a single
 query.\ [#]_
+
+You can optionally set the ``Accept`` header to control the
+``application/vnd.dwave.sapi.problems+json; version=<x>`` resource
+representation.\ [#v3]_
 
 .. dropdown:: Fields in Request Body
 
@@ -682,7 +688,27 @@ query.\ [#]_
         params          Solver-specific
                         :ref:`hybrid parameters <opt_index_properties_parameters>`
                         or :ref:`QPU parameters <qpu_solver_parameters>`.
-        solver          Solver to be used.
+        solver          Unique ID (name) of the solver to be used. For resource
+                        representation ``version=3``, the solver name is
+                        represented by the following JSON structure::
+
+                        "solver": {"name": solver_name, "version": {"graph_id": graph_id}
+
+                        where the fields are as follows:
+                        
+                        *    ``solver``: Parent for solver-related fields.
+                        *    ``name``: Name of the solver.
+                        *    ``version``: Parent for QPU-version-related fields.
+                             This field does not apply to hybrid solvers.
+                        *   ``graph_id``: ID of the QPU solver's
+                            :ref:`working graph <topologies_working_graph>`.
+                            This field does not apply to hybrid solvers.
+                        
+                        For resource representation ``version=2`` and
+                        earlier, simply use the name of the solver, including
+                        its version number. The ``version=2`` resource
+                        representation is deprecated; instead, use the
+                        ``version=3`` JSON structure.
         type            One of the supported
                         :ref:`problem types <property_qpu_supported_problem_types>`
                         (on QPUs, see the equivalent property for hybrid solvers
@@ -791,8 +817,12 @@ sampler,see the :ref:`sapi_rest_full_examples` section.
         .. doctest:: rest_live
             :skipif: test_api_token_set == False or hss_spot_check == False
 
+            >>> session = requests.Session()
+            >>> session.headers = {'X-Auth-Token': SAPI_TOKEN,
+            ...         'Content-type': 'application/json',
+            ...         'Accept': 'application/vnd.dwave.sapi.problems+json; version=3.0.0'}
             >>> r = session.post(f"{SAPI_HOME}/problems",
-            ...                   json=[{"solver": solver_hybrid_bqm,
+            ...                   json=[{"solver": {"name": bqm_solver},
             ...                          "label": "REST Submission to hybrid BQM solver",
             ...                          "data": {"format": "ref", "data": problem_data_id},
             ...                          "type": "bqm",
@@ -803,7 +833,7 @@ sampler,see the :ref:`sapi_rest_full_examples` section.
         This example sets the identifier, ``problem_data_id``, of the problem
         data being uploaded to the value returned from the
         :ref:`sapi_rest_get_multi_id` example and sets the requested solver to
-        ``hybrid_binary_quadratic_model_version2p``, a solver available to the
+        ``hybrid_binary_quadratic_model_version2``, a solver available to the
         user account that executed this example.
 
         .. code-block:: bash
@@ -811,12 +841,13 @@ sampler,see the :ref:`sapi_rest_full_examples` section.
             $ problem_data_id="be806ff1-09d3-49d4-bba1-3008790c99d6"
             $ curl -H "X-Auth-Token: $SAPI_TOKEN" $SAPI_HOME/problems -X POST \
             -d '[{"type":"bqm","label": "REST Submission to hybrid BQM solver", \
-            "solver":"hybrid_binary_quadratic_model_version2p", \
+            "solver":"hybrid_binary_quadratic_model_version2", \
             "data":{"format": "ref", "data": "'"$problem_data_id"'"}, \
             "params":{"time_limit":5}}]'
 
-.. [#]
-   |resource_rep_version|
+.. [#v3]
+    ``<x>`` is the version of the resource representation that the client
+    prefers; valid values are ``2`` (the default) and ``3``.
 
 .. [#]
     The request body is JSON data enclosed in a list. For submitting multiple
@@ -833,7 +864,7 @@ sampler,see the :ref:`sapi_rest_full_examples` section.
     >>> r.json()   # doctest: +SKIP
     {'status': 'IN_PROGRESS',
      'id': 'a239c000-10de-4663-b562-af8029c0c470',
-     'solver': 'hybrid_binary_quadratic_model_version2p',
+     'solver': {'name': 'hybrid_binary_quadratic_model_version2'},
      'type': 'bqm',
      'submitted_on': '2022-09-01T16:48:36.052206Z',
      'label': 'REST Submission to hybrid BQM solver'}
@@ -841,7 +872,8 @@ sampler,see the :ref:`sapi_rest_full_examples` section.
     >>> r.json()  # doctest: +SKIP
     {'status': 'COMPLETED',
      'id': '51fa2c98-3f19-4dd9-9f2d-c4295278218f',
-     'solver': 'Advantage_system4.1',
+     'solver': {'name': 'Advantage_system4.1',
+        'version': {'graph_id': '01d07086e1'}},
      'type': 'ising',
      'submitted_on': '2022-09-01T17:55:00.877038Z',
      'label': 'QPU REST submission 1',
@@ -893,15 +925,17 @@ Cancel Multiple Problems
 ------------------------
 
 To cancel pending problems (problems with status ``PENDING``), send an HTTP
-DELETE request to ``problems`` and, optionally, the
-``application/vnd.dwave.sapi.problems+json; version=<x>`` resource
-representation.\ [#]_
+DELETE request to the ``problems`` endpoint.
 
 The request body should be a JSON-encoded list of problem IDs; if the request
 body is empty, the request has no effect.
 
 When possible, if you have more than one problem to cancel, submit them in a
 single query.
+
+You can optionally set the ``Accept`` header to control the
+``application/vnd.dwave.sapi.problems+json; version=<x>`` resource
+representation.\ [#v3]_
 
 .. tab-set::
 
@@ -922,9 +956,6 @@ single query.
             $ id_list=[\"74d9344c-0160-47bc-b7b1-e245b2ffd955\",\"25f98470-bc55-476c-9042-120bbc0336cf\"]
             $ curl -H "X-Auth-Token: $SAPI_TOKEN" $SAPI_HOME/problems -X DELETE -d $id_list
 
-.. [#]
-    |resource_rep_version|
-
 .. dropdown:: 2xx responses
     :color: success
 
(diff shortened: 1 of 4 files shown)


docs/shared/sapi_rest.rst
```diff
@@ -91,11 +91,36 @@ The serialized file may look like this:
     ============== ==============================
     avg_load       Average current load for the solver.
     description    Description of the solver.
-    id             Unique ID (name) of the solver.
+    id             Unique ID (name) of the solver. Only available for an
+                   ``Accept`` request header  with ``version=2.0.0``, which is
+                   the default.
```
randomir, line 96: Representation version varies per resource. Some are 2.0, some 2.1, etc, but they are all currently 2.x. In the future they'll be 3.x.

Also, we should refer to "resource representation version", not the accept header:

```suggestion
    id             Unique ID (name) of the solver. Only available in
                   solver representation versions 1 and 2.
```

docs/shared/sapi_rest.rst
```diff
@@ -91,11 +91,36 @@ The serialized file may look like this:
...
+                   an ``Accept`` request header with ``version=3.0.0``.
+
+                   .. code-block:: json
+                    
+                        "identity":{
+                            "name":
+                            "version":{
+                                "graph_id":
+                            }
+                        }
```
randomir, line 111:
```suggestion
                        "identity": {
                            "name": <name>,
                            "version": {
                                "graph_id": <graph-id>
                            }
                        }
```

docs/shared/sapi_rest.rst
```diff
@@ -91,11 +91,36 @@ The serialized file may look like this:
...
+                    
+                   where the fields are as follows:
+                   
+                   *    ``identity``: Parent for solver-related fields.
+                   
+                   *    ``name``: Name of the solver.
+
+                   *    ``version``: Parent for version-related fields.
```
randomir, line 119: Version too is not relevant/present for hybrid solvers.

docs/leap_sapi/sapi_rest.rst
```diff
@@ -163,7 +163,45 @@ resources and provides example usages.
...
+/problems/<problem_id>/messages/             GET    application/vnd.dwave.sapi.problem-message+json; version=<x>
+/solvers/remote/                             GET    application/vnd.dwave.sapi.solver-definition-list+json; version=<x>
+/solvers/remote/<solver_id>/                 GET    application/vnd.dwave.sapi.solver-definition+json; version=<x> [#]_
+/solvers/remote/<solver_name>                GET    application/vnd.dwave.sapi.solver-definition+json; version=<x> [#]_
+============================================ ====== ===================================================================
+
+``<x>`` is the version of the Accept request-header; for example, ``3.0.0``.
+
+.. [#]
+    If ``version=3.0.0`` is specified, the ``identity``
```
randomir, line 195: I wouldn't be so specific in the docs, saying it has to be `3.0.0`. Both `3` and `3.0` work, so I would use `version=3`.

Also, it's important to note that this is not the API version per se, rather, it's the resource representation version that client prefers. It's at server's discretion to respond with the best representation. For example, if client asks for v3.1, the server might respond with 3.0.0 (if 3.1 is not available). But also, if it asks for 2.0 or 4.0 after the deprecation period, the server will respond with the latest v3 available.

docs/leap_sapi/sapi_rest.rst
```diff
@@ -163,7 +163,45 @@ resources and provides example usages.
...
+    resource field is returned; otherwise, the ``id`` resource field is
+    returned. For more information, see the "Solver Resource Fields"
+    table on the "2xx responses" tab in
+    :ref:`sapi_rest_get_remote_solver_config`.
+
+.. [#]
+    ``solver_name`` is the value in the ``identity.name`` field.
+    Regardless of the version specified in ``version``, the ``identity``
+    resource field is always returned.
```
randomir, line 204: This is unclear at best, false at worst. :slightly_smiling_face:

docs/leap_sapi/sapi_rest.rst
```diff
@@ -163,7 +163,45 @@ resources and provides example usages.
 Examples Setup
 --------------
 
-The examples below use the following setup:
+To run the examples in this topic, set up your development environment, such as
+iPython, as described on the following tabs.
```
randomir, line 167: it's `IPython`, but I'm not sure we need to call out one Python interpreter shell specifically.

docs/quantum_research/topologies.rst
```diff
@@ -32,10 +32,22 @@ The Working Graph
...
 
 Some small number of qubits and couplers in a QPU may not meet the
-specifications to function as desired. These are therefore removed from the
-programmable fabric that users can access. The subset of the QPU's graph
-available to users is the *working graph*. The yield\ [#]_ of the working graph
-is the percentage of working qubits that are present.
+specifications to function as desired. These qubits and couplers are therefore
+removed from the programmable fabric that users can access. The subset of the
+QPU's graph available to users is the *working graph*. The yield\ [#]_ of the
+working graph is the percentage of working qubits that are present.
```
randomir, line 38: Isn't the yield is a QPU/processor/wafer property, not working graph's property?

docs/quantum_research/topologies.rst
```diff
@@ -32,10 +32,22 @@ The Working Graph
...
+
+Due to various events during the operation of the QPU, qubits may no longer meet
+the required specifications and at that time the qubits and associated couplers
```
randomir, line 41: This feels repetitive after the previous paragraph.

docs/quantum_research/topologies.rst
```diff
@@ -32,10 +32,22 @@ The Working Graph
...
+are removed from the working graph; qubits and associated couplers may also be
+added back. Both adding and removing qubits and couplers result in the working
+graph being versioned. The version of the working graph is tracked in the
+``graph_id`` solver resource field (see the "Solver Resource Fields" table on
```
randomir, line 45: Perhaps also mention that previously we versioned the working graph through solver name/ID (i.e. version in the ID).

docs/shared/sapi_rest.rst
```diff
@@ -91,11 +91,36 @@ The serialized file may look like this:
...
-    id             Unique ID (name) of the solver.
+    id             Unique ID (name) of the solver. Only available for an
+                   ``Accept`` request header  with ``version=2.0.0``, which is
+                   the default.
     properties     :ref:`Solver properties <qpu_solver_properties_all>`
                    that reside in the ``properties`` dict; for example,
                    supported problem types, active qubits, active couplers,
                    total number of qubits, and so on.
+    identity       Represents the following JSON structure. Only available for
+                   an ``Accept`` request header with ``version=3.0.0``.
```
randomir, line 102: Same comment as above.

randomir (commented):
Quick first pass.

docs/quantum_research/topologies.rst
JoelPasvolsky: I think this is a problem of ambiguous wording (that "of") for the yield associated with the working graph or the working graph's yield.

docs/quantum_research/topologies.rst
```diff
@@ -32,10 +32,22 @@ The Working Graph
...
-specifications to function as desired. These are therefore removed from the
-programmable fabric that users can access. The subset of the QPU's graph
-available to users is the *working graph*. The yield\ [#]_ of the working graph
-is the percentage of working qubits that are present.
+specifications to function as desired. These qubits and couplers are therefore
+removed from the programmable fabric that users can access. The subset of the
+QPU's graph available to users is the *working graph*. The yield\ [#]_ of the
+working graph is the percentage of working qubits that are present.
+
+Due to various events during the operation of the QPU, qubits may no longer meet
```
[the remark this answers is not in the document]
randomir: `identity` will be a solver object attribute, like `name` was. `graph_id` will be a "derived property", also accessible directly on the solver object (`dwave.cloud.solver.Solver`).

docs/leap_sapi/sapi_rest.rst
```diff
@@ -163,7 +163,45 @@ resources and provides example usages.
...
+
+The ``Accept`` request header is optional; if included, it must be one of the
```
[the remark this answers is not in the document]
randomir: @JoelPasvolsky, Content types (with versions) are not new, but are lacking in the current documentation. They should be documented, but under each route individually, IMO, not in a table. Together with the current (latest) version, and all supported versions (2 and 3 after launch; then only 3 after sunset date).

docs/leap_sapi/sapi_rest.rst
charleswhchan: also note: only some of endpoints have new version (the ones affected)
so instead of including a list, maybe better to include the value under each endpoint?

docs/quantum_research/topologies.rst
qwriter: This sentence calls out that qubits may need to be removed/added after the initial calibration.

docs/quantum_research/topologies.rst
qwriter: I will include this in the description for the ``id`` solver resource field.

docs/quantum_research/topologies.rst
qwriter: Reworded and moved to a footnote.

docs/quantum_research/topologies.rst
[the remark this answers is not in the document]
qwriter: Clarified description of yield and moved to a footnote.

docs/leap_sapi/sapi_rest.rst
```diff
@@ -1038,6 +1093,9 @@ has completed processing.
 You can filter the results using one or more optional filter keys, concatenated
 with an ampersand ("&").
 
+You can optionally set the ``Accept`` header to
+``application/vnd.dwave.sapi.problems+json; version=<x>``.\ [#v3]_
```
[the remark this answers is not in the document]
randomir: I agree this is a decent placement, but not very useful.

1. We should significantly expand the "SAPI Version" section, to explain SAPI versioning in the first place.
2. We should explicitly, and in detail, explain what versions or resource representation are available *for each endpoint*, and how responses look/differ.

I believe, but please double-check, HSS endpoints (for multi-part upload, etc) are still at version 2, and don't support v3. In that case, we should say that explicitly as well -- only v2 is supported.

To answer Joel's question about how much will this be used -- it doesn't matter, we have to document it in SAPI docs. Anyone actually reading this page will yearn to know about representation versions!

On that note, we should also say next to v2 it's deprecated, and it will be removed in 6 months.

docs/quantum_research/topologies.rst
```diff
@@ -51,6 +52,13 @@ is the percentage of working qubits that are present.
     commercial systems typically have more than 97% of fabricated qubits
     available in their working graphs.
 
+    The version of the working graph is tracked in the ``graph_id`` solver
+    resource field (see
+    :ref:`solver properties <https://docs.dwavequantum.com/en/latest/ocean/api_ref_cloud/solver.html#properties>`).
```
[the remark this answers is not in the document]
qwriter: Corrected xref.

docs/leap_sapi/sapi_rest.rst
```diff
@@ -263,7 +268,9 @@ Initiate Upload of Problem Data
 -------------------------------
 
 For quantum-classical hybrid solvers in the Leap service, you must upload your
-problem using the ``$SAPI_HOME/bqm/multipart/*`` endpoints before submitting a
+problem using the ``$SAPI_HOME/bqm/multipart/*`` endpoints and, optionally,
```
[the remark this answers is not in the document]
qwriter: Creating tabs for these instead per this [comment](https://github.com/dwavesystems/dwave-ocean-sdk/pull/369#discussion_r2283168305).

docs/leap_sapi/sapi_rest.rst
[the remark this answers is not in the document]
qwriter: I will create another PR for "We should significantly expand the SAPI Version section, to explain SAPI versioning in the first place."

docs/leap_sapi/sapi_rest.rst
```diff
@@ -1933,7 +2119,6 @@ data. Calculations of the checksums for the part(s) and whole are shown there.
 
 .. doctest:: rest_live
     :skipif: test_api_token_set == False or hss_spot_check == False
-    :hide:
 
     >>> # blocking call to sequence r6 (part upload) and r7 (combine checksum)
     >>> r6.json()
```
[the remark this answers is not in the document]
qwriter: But when we try to run the example in the "Retrieve Your Solutions" section, then there is no solution because the``combine`` wasn't run.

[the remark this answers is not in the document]
JoelPasvolsky: This part is just to prevent automated tests from failing because they happen faster than an interaction with a user:

image

docs/leap_sapi/sapi_rest.rst
```diff
@@ -1669,7 +1768,9 @@ used for all SAPI requests.
     >>> SAPI_TOKEN = "ABC-1234567...345678"     # doctest: +SKIP
     ...
     >>> session = requests.Session()
-    >>> session.headers = {'X-Auth-Token': SAPI_TOKEN, 'Content-type': 'application/json'}
+    >>> session.headers = {'X-Auth-Token': SAPI_TOKEN,
+    ...                    'Content-type': 'application/json',
+    ...                    'Accept': 'application/vnd.dwave.sapi.solver-definition-list+json; version=3.0.0'}
```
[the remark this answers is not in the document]
qwriter: But this is the initial setup for the end-to-end examples section.

docs/leap_sapi/sapi_rest.rst
charleswhchan, line 1: as far as I can tell, all of the Accept: problem/s headers are on version=2.1 atm

randomir: Actually, it's version 2.0.0 for `/solver/` endpoints, and 2.1.0 for `/problems/` endpoints.

docs/shared/sapi_rest.rst
```diff
@@ -91,7 +111,34 @@ The serialized file may look like this:
...
+                   
+                   *    ``identity``: Parent for solver-related fields.
+                   
+                   *    ``name``: Name of the solver.
+
+                   *    ``version``: Parent for QPU-version-related fields. This
+                        field does not apply to hybrid solvers.
+                   
+                   *    ``graph_id``: ID of the QPU solver's
+                        :ref:`working graph <topologies_working_graph>`.
```
randomir, line 141:
```suggestion
    identity       Represents the following JSON object. Only available for
                   resource representation ``version=3``.

                   .. code-block:: json
                    
                        {
                            "name": name
                            "version": {
                                "graph_id": graph_id
                            }
                        }
                    
                   where the values are as follows:
                   
                   *    ``name``: Name of the solver.

                   *    ``version``: Version object, used for QPU solvers only.
                   
                   *    ``graph_id``: ID of the QPU solver's
                        :ref:`working graph <topologies_working_graph>`.

```

docs/shared/sapi_rest.rst
```diff
@@ -72,6 +72,26 @@ The serialized file may look like this:
...
+                    *    ``name``: Name of the solver.
+                    *    ``version``: Parent for QPU-version-related fields.
+                         This field does not apply to hybrid solvers.
+                    *   ``graph_id``: Identifier of the QPU solver's
+                        :ref:`working graph <topologies_working_graph>`.
+
+                    For resource representation ``version=2`` and
+                    earlier, the solver name is represented as a string.
+                    The ``version=2`` resource representation is deprecated;
+                    instead, use the ``version=3`` JSON structure.
```
randomir, line 94: I pointed out this in the first review as well. We can't have an incomplete JSON value here. Check the [specs](https://www.json.org/json-en.html) for ref.
```suggestion
    solver          Unique identifier (name) of the solver. For resource representation
                    ``version=3``, the solver is represented by the
                    following JSON object::

                    {"name": name, "version": {"graph_id": graph_id}}

                    where the fields are as follows:
                   
                    *    ``name``: Name of the solver.
                    *    ``version``: Version object, used for QPU solvers only.
                    *   ``graph_id``: Identifier of the QPU solver's
                        :ref:`working graph <topologies_working_graph>`.

                    For resource representation ``version=2`` and
                    earlier, the solver is identified with a string.
                    The ``version=2`` resource representation is deprecated;
                    instead, use the ``version=3`` JSON structure.
```

randomir (commented):
Generally looks very nice, but there are still a few details we need to fix before merging.

docs/leap_sapi/sapi_rest.rst
```diff
@@ -1309,6 +1452,12 @@ The request should contain no body.
             $ problem_id="74d9344c-0160-47bc-b7b1-e245b2ffd955"
             $ curl -H "X-Auth-Token: $SAPI_TOKEN" $SAPI_HOME/problems/$problem_id/answer -X GET
 
+.. dropdown:: Optional ``Accept`` header
+
+    SAPI supports the resource representation:
+
+    *   ``application/vnd.dwave.sapi.problem-answer+json; version=2.1``
```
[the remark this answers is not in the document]
randomir: @charleswhchan, @cvillamizarag, thoughts? For both the answer and messages endpoint.

docs/leap_sapi/sapi_rest.rst
```diff
@@ -667,7 +693,29 @@ query.\ [#]_
...
+                        *    ``version``: Parent for QPU-version-related fields.
+                             This field does not apply to hybrid solvers.
+                        *   ``graph_id``: Unique identifier of the QPU solver's
+                            :ref:`working graph <topologies_working_graph>`.
+
+                        For resource representation ``version=2`` and
+                        earlier, use the name of the solver. The ``version=2``
+                        resource representation is deprecated; instead, use the
+                        ``version=3`` JSON structure.
+
```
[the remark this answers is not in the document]
qwriter: > should we include an example
> 
> ```
> "solver": "solver_id"
> ```

@charleswhchan -- Just to double-check, this is for `version=3`, correct?

docs/leap_sapi/sapi_rest.rst
```diff
@@ -767,14 +816,15 @@ sampler,see the :ref:`sapi_rest_full_examples` section.
...
+        ``bqm_solver`` selected as in
         the :ref:`sapi_rest_quick_start_submit_hybrid` section.
 
         .. doctest:: rest_live
             :skipif: test_api_token_set == False or hss_spot_check == False
 
+            >>> session.headers['Accept'] = 'application/vnd.dwave.sapi.problems+json; version=3'
             >>> r = session.post(f"{SAPI_HOME}/problems",
-            ...                   json=[{"solver": solver_hybrid_bqm,
+            ...                   json=[{"solver": bqm_solver,
```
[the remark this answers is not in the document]
qwriter: This change resulted in the removal of the associated `testsetup` directive.

docs/leap_sapi/sapi_rest.rst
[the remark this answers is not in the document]
charleswhchan: @qwriter yes

charleswhchan (approved):

qwriter:
@charleswhchan -- In your [comment](https://github.com/dwavesystems/dwave-ocean-sdk/pull/369#discussion_r2323715135), do you mean that we should add a note: ""Routes only accepting v2 should return error when version is not supported."

Would this apply to all `version=2.1`?

docs/leap_sapi/sapi_rest.rst
```diff
@@ -667,7 +693,27 @@ query.\ [#]_
...
+                        deprecated; instead, use the ``version=3`` JSON
+                        structure. For resource representation ``version=3``,
+                        the solver identifier is represented by the following
+                        JSON structure::
+
+                        {"name": solver_name, "version": {"graph_id": graph_id}}
+
+                        where the fields are as follows:
+
+                        *    ``name``: Name of the solver.
```
randomir, line 709: Either change this to ``solver_name``, or change the above JSON to use ``name`` for solver name.

docs/leap_sapi/sapi_rest.rst
```diff
@@ -1053,8 +1148,9 @@ with an ampersand ("&").
                         if not specified.
         status          Problem state: COMPLETED, IN_PROGRESS, PENDING, FAILED,
                         CANCELLED.
-        solver          Solver name (e.g.,
-                        ``hybrid_binary_quadratic_model_version2p``).
+        solver          Solver name as a string (e.g.,
+                        ``hybrid_binary_quadratic_model_version2``) or a dict
+                        (e.g, ``'solver': {'name': 'hybrid_binary_quadratic_model_version2'}``).
```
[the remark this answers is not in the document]
randomir: My comment above is still relevant. Dict can't be used for solver on the problems endpoint.

docs/leap_sapi/sapi_rest.rst
```diff
@@ -1146,15 +1249,29 @@ not completed processing.
...
 
     .. tab-item:: cURL
 
         .. code-block:: bash
 
             $ problem_id="74d9344c-0160-47bc-b7b1-e245b2ffd955"
-            $ curl -H "X-Auth-Token: $SAPI_TOKEN" $SAPI_HOME/problems/$problem_id?timeout=5 -X GET
+            $ auth="X-Auth-Token: $SAPI_TOKEN"
+            $ accept="Accept: application/vnd.dwave.sapi.problem+json; version=3"
+            $ curl -H "$auth" -H "$accept" $SAPI_HOME/problems/$problem_id?timeout=5 -X GET
```
randomir, line 1262: You can remove `-X GET` for simplicity. GET is the default verb used.

docs/leap_sapi/sapi_rest.rst
```diff
@@ -1221,15 +1338,29 @@ The request should contain no body.
...
 
     .. tab-item:: cURL
 
         .. code-block:: bash
 
             $ problem_id="74d9344c-0160-47bc-b7b1-e245b2ffd955"
-            $ curl -H "X-Auth-Token: $SAPI_TOKEN" $SAPI_HOME/problems/$problem_id/info -X GET
+            $ auth="X-Auth-Token: $SAPI_TOKEN"
+            $ accept="Accept: application/vnd.dwave.sapi.problem-data+json; version=3"
+            $ curl -H "$auth" -H "$accept" $SAPI_HOME/problems/$problem_id/info -X GET
```
randomir, line 1351: same comment as above:
```suggestion
            $ curl -H "$auth" -H "$accept" $SAPI_HOME/problems/$problem_id/info
```

docs/leap_sapi/sapi_rest.rst
```diff
@@ -1464,19 +1619,37 @@ to get a subset of solver fields.
         .. doctest:: rest_live
             :skipif: test_api_token_set == False
 
-            >>> filter = urlencode({"filter": "none,+id,+status,+avg_load,+properties.num_qubits,+properties.category"})
+            >>> session.headers['Accept'] = 'application/vnd.dwave.sapi.solver-definition-list+json; version=3'
+            >>> filter = urlencode({"filter": "none,+identity,+status,+avg_load"})
             ...
             >>> r = session.get(f"{SAPI_HOME}/solvers/remote/?{filter}")
```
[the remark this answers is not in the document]
randomir: Any reason you prefer the use of `urlencode` here?

docs/leap_sapi/sapi_rest.rst
```diff
@@ -1684,9 +1859,11 @@ quantity of retrieved information, can be omitted.
 .. doctest:: rest_live
     :skipif: test_api_token_set == False
 
-    >>> filter = urlencode({"filter": "none,+id,+status,+avg_load,+properties.num_qubits,+properties.category"})
+    >>> params = {"filter": "none,+identity,+status,+avg_load,+properties.num_qubits,+properties.category"}
     ...
-    >>> r1 = session.get(f"{SAPI_HOME}/solvers/remote/?{filter}")
+    >>> r1 = requests.get(f"{SAPI_HOME}/solvers/remote/", params=params,
+    ...                   headers={'X-Auth-Token': SAPI_TOKEN,
```
randomir, line 1865: Isn't `X-Auth-Token` already set globally on the session?

randomir (commented):
Looking good, just a few comments this time.

charleswhchan:
> @charleswhchan -- In your [comment](https://github.com/dwavesystems/dwave-ocean-sdk/pull/369#discussion_r2323715135), do you mean that we should add a note: ""Routes only accepting v2 should return error when version is not supported."
> 
> Would this apply to all `version=2.1`?

No doc change.

Basically, I think the issue Rad pointed out is that for API endpoints, we should only `Accept` version supported and nothing more. However, some endpoints are NOT honoring this contract so it is considered a bug. (Filed LWS-862)

docs/leap_sapi/sapi_rest.rst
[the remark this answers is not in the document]
randomir: @charleswhchan, on closer inspection, I see the endpoints in question actually return correct version (2.1), even when 3.0 is asked for.

IMO, this is the correct behavior, the one we actually have in the technical design now (if version client asked for is not supported, just return the latest). In this case client asks for v3, and the server returns the latest, which is v2.

The bug is in the technical design, where version is bumped to v3 on all endpoints, even the ones that didn't change (and still return v2). So, no code changes needed, just docs.

docs/leap_sapi/sapi_rest.rst
JoelPasvolsky, line 709: lines 709 to 711 are indented to column 30 instead of 29

JoelPasvolsky: > Either change this to `solver_name`, or change the above JSON to use `name` for solver name.

Also notice that line 698 has this same value as `solver_id`

docs/leap_sapi/sapi_rest.rst
```diff
@@ -744,37 +790,27 @@ query.\ [#]_
                            Not used for ``data`` with ``format="ref"``.
         ================== =======================================
 
-For an example of sumitting a problem to a quantum processing unit (QPU)
+For an example of submitting a problem to a quantum processing unit (QPU)
 sampler,see the :ref:`sapi_rest_full_examples` section.
 
-.. testsetup:: rest_live
```
JoelPasvolsky, line 750: I see you deleted the test setup here and switched from a hybrid solver returned from Leap to a hardcoded one in line 813. Currently we're not running live tests from CI but if we start then a change to ``hybrid_binary_quadratic_model_version2`` will cause the example to fail. 
Maybe we can live fine with this.

docs/leap_sapi/sapi_rest.rst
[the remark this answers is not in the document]
randomir: That's exactly it, @JoelPasvolsky. See also [proactive content negotiation](https://datatracker.ietf.org/doc/html/rfc9110#name-proactive-negotiation):
> A user agent cannot rely on proactive negotiation preferences being consistently honored, since the origin server might not implement proactive negotiation for the requested resource or might decide that sending a response that doesn't conform to the user agent's preferences is better than sending a [406 (Not Acceptable)](https://datatracker.ietf.org/doc/html/rfc9110#status.406) response.

TL;DR: the client sends its wishes, and the server responds with whatever representation it sees fit. In our case, that's always the latest version. (Except during the deprecation period when both v2 and v3 are supported.)

This approach gives SAPI a lot of flexibility, as it's up to the client to verify if the version returned can be processed. In Ocean, we verify version returned using semver operators like `~=2.1` (meaning `>=2.1,<3`), and fail hard if version is not supported.

This also means per-resource versions can easily be added and removed.

docs/leap_sapi/sapi_rest.rst
```diff
@@ -812,15 +865,16 @@ sampler,see the :ref:`sapi_rest_full_examples` section.
     >>> r.json()   # doctest: +SKIP
```
JoelPasvolsky, line 865: A couple of things are unclear to me here: 

1. "example 201 responses" --> but I'm seeing 200, not 201.
```
>>> r
<Response [200]>
```
2. I'm seeing the JSON response wrapped as a list:
```
>>> r.json()
[{'status': 'PENDING',
  'id': '306f60a ...
sion 2'}]
```

charleswhchan (approved):

docs/leap_sapi/sapi_rest.rst
JoelPasvolsky: Line 698, "For resource representation ``version=2.1`` and earlier, use the name of the solver (e.g., ``"solver": "solver_id"``)", means the user would set `"solver": "Advantage_system4.1"`, for example; Line 709 would be `{"name": "Advantage_system4.1", "version": {"graph_id": "01d07086e1"}}`, for example. 
So for the same value, "Advantage_system4.1", in 698 you have `solver_id` and in 709 you have `name`

docs/leap_sapi/sapi_rest.rst
JoelPasvolsky: See [my previous comment](https://github.com/dwavesystems/dwave-ocean-sdk/pull/369#discussion_r2283187370) about setting just the `accept` for the existing session iInstead of setting `headers` from scratch in the request:

```
session.headers['Accept'] = 'application/vnd.dwave.sapi.solver-definition-list+json; version=3'
r1 = requests.get(f"{SAPI_HOME}/solvers/remote/", params=params)
```

qwriter: hmm... that doesn't seem to work either. 
image

randomir:
> @randomir -- See @JoelPasvolsky's [comment](https://github.com/dwavesystems/dwave-ocean-sdk/pull/369#discussion_r2325717467).

@qwriter, see my answer [here](https://github.com/dwavesystems/dwave-ocean-sdk/pull/369/files#r2325802581).

docs/leap_sapi/sapi_rest.rst
```diff
@@ -1607,9 +1607,8 @@ to get a subset of solver fields.
             :skipif: test_api_token_set == False
 
             >>> params = {"filter": "none,+identity,+status,+avg_load"}
-            >>> r = requests.get(f"{SAPI_HOME}/solvers/remote/", params=params,
-            ...                  headers={'X-Auth-Token': SAPI_TOKEN,
-            ...                  'Accept': 'application/vnd.dwave.sapi.solver-definition-list+json; version=3'})
+            >>> r = session.get(f"{SAPI_HOME}/solvers/remote/", params=params,
+            ...                  headers={'Accept': 'application/vnd.dwave.sapi.solver-definition-list+json; version=3'})
```
JoelPasvolsky, line 1611: This is one of only two places where you are setting the `Accept` header inside the GET. In ten places we prefer the `>>> session.headers['Accept'] =...` construct, because it's easier for users to parse.
Why not here too?

docs/leap_sapi/sapi_rest.rst
```diff
@@ -1607,9 +1607,8 @@ to get a subset of solver fields.
...
```
JoelPasvolsky, line 1612: Trailing spaces

docs/leap_sapi/sapi_rest.rst
```diff
@@ -1859,9 +1858,8 @@ quantity of retrieved information, can be omitted.
 
     >>> params = {"filter": "none,+identity,+status,+avg_load,+properties.num_qubits,+properties.category"}
     ...
-    >>> r1 = requests.get(f"{SAPI_HOME}/solvers/remote/", params=params,
-    ...                   headers={'X-Auth-Token': SAPI_TOKEN,
-    ...                   'Accept': 'application/vnd.dwave.sapi.solver-definition-list+json; version=3'})
+    >>> r1 = session.get(f"{SAPI_HOME}/solvers/remote/", params=params,
+    ...                   headers={'Accept': 'application/vnd.dwave.sapi.solver-definition-list+json; version=3'})
```
JoelPasvolsky, line 1862: Same comment about placement of `Accept` header

JoelPasvolsky (approved):

PR #371, opened 2025-08-07
Update properties for Advantage2 recalibration

qwriter:
Qubit and coupler counts were confirmed by PD.


JoelPasvolsky (approved):
Matches what I'm seeing online now, thanks.

PR #383, opened 2025-09-11
Ensure CQM example returns feasible results & update statement on Zephyr

JoelPasvolsky:
Two minor fixes (DOC-890 & DOC-998): CQM example did not always return feasible result as the first sample and Zephyr-based QPU is now a product. 
Related: [PR-596](https://github.com/dwavesystems/dwave-system/pull/596)


qwriter:
LGTM!

PR #389, opened 2025-09-26
Add data description to SAPI REST Interface topic

qwriter:


docs/leap_sapi/sapi_rest.rst
```diff
@@ -787,6 +787,12 @@ query.\ [#]_
         ================== =======================================
         Key                Value and Encoding
         ================== =======================================
+        data               An identifier, of type
```
JoelPasvolsky, line 790:
```suggestion
        data               For ``data`` with ``format="ref"``, an identifier, of type
```

For consistency with the ``lin`` and ``quad`` field descriptions. We're using this format here to place optional fields in the correct usage.

JoelPasvolsky (approved):

PR #464, opened 2026-06-04
Update sitemap & canonical tags

JoelPasvolsky:
#### What does this implement/fix?
Sitemap will list all pages instead of just top page and adds canonical tags.

#### AI Generation Disclosure
No AI tools used though the request to make this update might have been based on AI recommendations.


arcondello (approved):
I am not sure I am qualified to review, but approved because it seems harmless.

qwriter (approved):
Builds locally. LGTM!

PR #467, opened 2026-06-12
Add the package version numbers to the docs

arcondello:
Feels useful to see the version of the package that's in the docs.

### Before
image

### After
image

#### AI Generation Disclosure
No AI used.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -301,6 +301,9 @@ def substitute_versions(app, docname, source):
     from dwave.system import __version__ as system_version
     source[0] = source[0].replace("|system_version|", system_version)
 
+    from dwaveoceansdk import __version__ as sdk_version
+    source[0] = source[0].replace("|sdk_version|", sdk_version)
+
     from minorminer import __version__ as minorminer_version
     source[0] = source[0].replace("|minorminer_version|", minorminer_version)
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -34,7 +34,6 @@ copyright = 'D-Wave'
 
 rst_prolog = f"""
 .. |python_requires| replace:: {python_requires}
-.. |sdk_version| replace:: {version}
 """
 
 # -- General configuration ------------------------------------------------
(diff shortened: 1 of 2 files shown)


thisac (approved):
Thanks @arcondello.

JoelPasvolsky (approved):
Thank you!

arcondello:
@thisac suggested adding them to the docs table as well

image

randomir (commented):
This is great, I would just make the version pop less 😄.

What I mean, for legibility, I would either add an emdash (`dimod -- 0.12.21`), or put version in parentheses (`dimod (0.12.21)`), or change the color (lighter shade), font (monospace?), size or style of the version number.

arcondello:
> What I mean, for legibility, I would either add an emdash (dimod -- 0.12.21), or put version in parentheses (dimod (0.12.21)), or change the color (lighter shade), font (monospace?), size or style of the version number.

To be clear, you mean on the sidebar or in the table or both?

randomir:
> To be clear, you mean on the sidebar or in the table or both?

Both. It doesn't have to be the same mechanism. I see emdash working in both places, perhaps even parens. In the sidebar italic would work as well, but probably easier to stick to content change vs style change.

arcondello:
With `—`
image

I think adding `—` to the table is just a bit too much punctuation  when combined with the `:` so I am inclined to leave it be?

Can push the commit if you prefer this. FWIW, I mildly prefer the original version but it's a very mild preference.

> probably easier to stick to content change vs style change.

Agree with this for now. Though we can always adjust later.

randomir:
The sidebar looks much better now, IMO.

For the table, have you tried with parentheses, like you have for "(deprecated)"?

randomir (approved):

arcondello:
image

Again, I kind of prefer the old one but fine to go with this if you like.

arcondello:
~Not sure why CI isn't firing for the latest commit :thinking: I don't see anything on the GH or CircleCI status pages.~ it eventually fired :shrug:

randomir:
Thanks @arcondello, this looks great!

image

PR #469, opened 2026-06-12
Build docs in parallel by default

arcondello:
This defaults to using the number of CPU cores. I think this is quite safe and speeds up builds locally and in CI.

#### AI Generation Disclosure
No AI used.


arcondello:
Sadly, looks like CircleCI doesn't like the parallelism.

PR #474, opened 2026-06-18
Move dwave-optimization images into a subdirectory

arcondello:
Annoyingly this doesn't actually fix the images in the package, because we install the version of PyPI and the docstrings there use the old path.

IMO it's OK to live with a few broken images until the next deploy.

#### AI Generation Disclosure
No AI used.


JoelPasvolsky:
> Annoyingly this doesn't actually fix the images

We could copy the images in temporarily until next release

arcondello:
>  We could copy the images in temporarily until next release

I'd rather deploy a package if we're that concerned. We're just very likely to forget to remove them later.

JoelPasvolsky:
This update add two warnings,
```
/root/project/docs/industrial_optimization/index.rst:: WARNING: image file not readable: _images/optimization_icon.svg [image.not_readable]
/root/project/docs/industrial_optimization/index.rst:: WARNING: image file not readable: _images/optimization_icon.svg [image.not_readable]
```
though I don't see immediately what is missing. 

Does updating the ``dwave-optimization`` package here do anything useful then, or should we take that out and leave that to be updated next release?

arcondello:
Oops, deleted one image I didn't mean to.

And yes, updating the submodule updates the `.rst` files but not the `.py` files.

JoelPasvolsky:
> And yes, updating the submodule updates the `.rst` files but not the `.py` files.

right, so it updates paths that aren't being used until the next sdk release, that's why i'm saying we could drop the ``dwave-optimization`` update from this PR for simplicity, but either way is okay, up to you

JoelPasvolsky (approved):

arcondello:
> right, so it updates paths that aren't being used until the next sdk release, 

Not quite. For instance the icons in the philosophy section are correctly updated here.

PR #487, opened 2026-08-27
Add rules for AI agents

randomir:
Added instruction files for AI coding agents. `AGENTS.md` provides repository-level guidance for this repo (with `CLAUDE.md` as a one-line bridge for Claude Code, which does not read `AGENTS.md` natively). `AGENTS.baseline.md` is the shared Ocean-wide baseline that individual Ocean package repositories reference, and `AGENTS.template.md` is the starter template for adding an `AGENTS.md` to an Ocean package repository. These files affect only AI-assisted development workflows; they have no effect on installed packages.

#### AI Generation Disclosure

Used Claude to simplify and extend my initial `CLAUDE.md`.


.agents/AGENTS.baseline.md
```diff
@@ -0,0 +1,60 @@
...
+
+- Prefer targeted reads (`grep`, `sed -n 'START,ENDp'`, symbol search) over
+  reading whole files.
+- Review your edits with `git diff -- <path>` before declaring done.
+
+## Python Defaults
+
+- Use `pip` and `python -m venv` for dependency and environment commands.
+- Use `logging` instead of `print()` in library code.
+- Add type hints to public functions, including return types.
```
qci-amos, line 36: I think it's reasonable to insist on type annotations everywhere

.agents/AGENTS.baseline.md
```diff
@@ -0,0 +1,60 @@
...
+- Never silently swallow exceptions: catch only exceptions you can
+  meaningfully handle, and let the rest propagate.
+- Ask before adding a runtime dependency. Ocean packages keep dependencies
+  minimal; new ones affect the whole SDK's install footprint.
+
+## Code Quality
+
+- Match the existing package style before introducing a new pattern; Ocean
```
qci-amos, line 44: what do you have in mind by "style"? is this a reference to black/ruff formatting or more abstract?

arcondello:
My first thought is that some of the contents are concerned with communication style rather than being Ocean specific. While I agree with the communication style personally, I am not sure it makes sense for Ocean's files to be opinionated on aspects of the agent interaction that will not effect the code changes.

.agents/AGENTS.baseline.md
```diff
@@ -0,0 +1,60 @@
...
+  personal preference.
+- Avoid speculative abstractions for single-use code.
+- Keep functions focused; prefer splitting once a function does more than
+  one thing (rough guide: ~50 lines).
+- After edits, run the package's lint command (see its `## Commands`) and
+  fix what it reports.
+
+## Commits and Pull Requests
+
+- When preparing a commit or PR, fetch and follow the full contributing
```
qci-amos, line 55: it's worth considering if we want to advise it to not do a commit or PR itself automatically and instead ask the human do to that step

.agents/AGENTS.template.md
qci-amos, line 1: Agents will NOT read this file?

qci-amos: this sort of file may be better suited to store elsewhere because I'm not clear if an agent would automatically find and try to do something with this

AGENTS.md
```diff
@@ -0,0 +1,32 @@
...
+
+Installer for D-Wave's Ocean SDK tools. The repo is mostly Ocean/Leap docs + package
+configuration that pins a set of dozen or so Ocean packages that contain the actual
+SDK code. Each package is in its repo, and added as a submodule here, mostly to
+facilitate the docs build.
+
+## Commands
+
+- Setup: `python -m venv .venv && pip install . && pip install -r tests/requirements.txt`
+- Test (all): `python -m unittest`
```
qci-amos, line 23: To what extent is `unittest` used/preferred/required in Ocean?

AGENTS.md
```diff
@@ -0,0 +1,32 @@
...
+## Project Overview
+
+Installer for D-Wave's Ocean SDK tools. The repo is mostly Ocean/Leap docs + package
+configuration that pins a set of dozen or so Ocean packages that contain the actual
+SDK code. Each package is in its repo, and added as a submodule here, mostly to
+facilitate the docs build.
+
+## Commands
+
+- Setup: `python -m venv .venv && pip install . && pip install -r tests/requirements.txt`
```
qci-amos, line 22: In my own usage, I've decided I really don't want agents touching my python installation and I don't have one venv-per-repo. I want to set that up myself and then they can use it...

.agents/AGENTS.baseline.md
```diff
@@ -0,0 +1,60 @@
...
+fetches this file on demand (see the per-package template).
+-->
+
+## Working Style
+
+- State your assumptions and a short plan of verifiable goals before editing.
+- If the task is ambiguous, ask at most one clarifying question; otherwise
+  proceed.
+- Make the smallest change that solves the problem; don't touch or refactor
+  unrelated code unless asked.
```
thisac, line 17: 💯

AGENTS.md
thisac: We use both in different packages. I believe I might have introduced further use of Pytest than was before. Generally I prefer `pytest`, since it can also run `unittest` tests, but I'm fine having both in use.

AGENTS.template.md
```diff
@@ -0,0 +1,58 @@
...
+
+<!-- 2–4 lines. What this package is and anything surprising about it.
+     Example (dimod-style): "Shared API for binary quadratic samplers.
+     Core is C++ under dimod/include/ with Cython bindings; files under
+     dimod/*.cpp are generated — edit the .pyx sources instead." -->
+
+## Commands
+
+<!-- Exact, copy-pasteable. Replace the examples with this package's real
+     invocations. This is the highest-value section in the file. -->
```
thisac:
```suggestion
<!-- Exact, copy-pasteable. Replace the examples with this package's real
     invocations, or remove if not used. This is the highest-value section in the file. -->
```

AGENTS.md
thisac: I don't think that I mind a venv being set up, but I wonder this could be opt-in instead, as in e.g. ask first. Could add in a separate `## Command execution guidelines` section below that details this.

```markdown
## Command execution guidelines

- Setup, Release note, Docs build: 
    Ask for permission before running these for the first time.
```

.agents/AGENTS.baseline.md
```diff
@@ -0,0 +1,60 @@
...
+- Before finishing, verify with the narrowest relevant test (see the
+  package's `## Commands` section for how to run them).
+
+## Context and Tool Use
+
+- Prefer targeted reads (`grep`, `sed -n 'START,ENDp'`, symbol search) over
+  reading whole files.
+- Review your edits with `git diff -- <path>` before declaring done.
+
+## Python Defaults
```
thisac, line 27: Look for installing dependency groups (preferred) if available, otherwise `requirements.txt`, and choose either `docs`, `tests`, `dev` groups/files depending on task. Not sure if this is too specific though.

.agents/AGENTS.baseline.md
```diff
@@ -0,0 +1,60 @@
...
+-->
+
+## Working Style
+
+- State your assumptions and a short plan of verifiable goals before editing.
+- If the task is ambiguous, ask at most one clarifying question; otherwise
+  proceed.
+- Make the smallest change that solves the problem; don't touch or refactor
+  unrelated code unless asked.
+- Before finishing, verify with the narrowest relevant test (see the
```
thisac, line 18: Potentially add something along the lines of
```markdown
- If you make a change beyond strict task necessity, ask before making any 
  changes and explicitly inform why it provides clear benefit.
```

.agents/AGENTS.baseline.md
thisac, line 27: Worth adding a reference to the current Python versions (e.g., in `pyproject.toml`) supported and make sure that all code written should be compatible with all active versions?

.agents/AGENTS.baseline.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Commits and Pull Requests
+
+- When preparing a commit or PR, fetch and follow the full contributing
+  guide:
+  https://raw.githubusercontent.com/dwavesystems/dwave-ocean-sdk/refs/heads/master/CONTRIBUTING.rst
+- The non-negotiables (listed in each package's AGENTS.md) always apply,
+  even if the guide was not fetched.
+
+## Communication
```
thisac, line 61: Maybe
```markdown
- Call out non-goals: what you intentionally did not change and why.
```

AGENTS.baseline.md
```diff
@@ -0,0 +1,60 @@
...
+## Code Quality
+
+- Match the existing package style before introducing a new pattern; Ocean
+  packages differ in age and conventions, so the surrounding code wins over
+  personal preference.
+- Avoid speculative abstractions for single-use code.
+- Keep functions focused; prefer splitting once a function does more than
+  one thing (rough guide: ~50 lines).
+- After edits, run the package's lint command (see its `## Commands`) and
+  fix what it reports.
```
thisac:
```suggestion
- After edits, run the package's lint command (see its `## Commands`), if there is one, and
  fix what it reports.
```

.agents/AGENTS.baseline.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Working Style
+
+- State your assumptions and a short plan of verifiable goals before editing.
+- If the task is ambiguous, ask at most one clarifying question; otherwise
+  proceed.
+- Make the smallest change that solves the problem; don't touch or refactor
+  unrelated code unless asked.
+- Before finishing, verify with the narrowest relevant test (see the
+  package's `## Commands` section for how to run them).
```
thisac, line 19: Worth adding
```markdown
- If one or more tests fail, try to fix the issue according to the guidelines 
  set out in this file, and rerun the same tests.
```

.agents/AGENTS.template.md
```diff
@@ -0,0 +1,58 @@
...
+
+- <e.g. "Public API is what's re-exported in <pkg>/__init__.py; underscore
+  modules are internal and not covered by semver.">
+
+## Non-negotiables (always apply)
+
+<!-- Copied from the Ocean baseline so they hold even when nothing is
+     fetched. Keep this list short and in sync with CONTRIBUTING.rst. -->
+
+- One focused feature/fix per PR.
```
thisac, line 47: Unless the code is handling PRs (which I think it shouldn't), will this do what we expect? Maybe formulating this along the lines of "Try to contain changes to a single focused feature/fix, or otherwise inform that that's not the case".

thisac (commented):
Thanks @randomir. I think this could be a nice addition. 

Minor, but can we move the agent files into an `agents` or `.agents` folder and still keep them discoverable? I'm not fond of having too many added files in root.

.agents/AGENTS.baseline.md
randomir: In PEP-8's [local consistency sense](https://peps.python.org/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds).

randomir:
> My first thought is that some of the contents are concerned with communication style rather than being Ocean specific. While I agree with the communication style personally, I am not sure it makes sense for Ocean's files to be opinionated on aspects of the agent interaction that will not effect the code changes.

Yes, a few of these are about efficient communication with the agent. I wrote `AGENTS.baseline.md` more like OSS/APT team standards than Ocean-specific rulebook. The reason for putting it here is simplicity. The alternative is to host them in a new (internal?) repo, together with hooks, skills, plugins and whatnot. Not saying we won't end up there, but I figured it's better to start with what we have right now.

.agents/AGENTS.baseline.md
randomir: This line does not imply automatic commit/push. I never let Claude do that for me, nor do I explicitly ask it to do it.

.agents/AGENTS.template.md
randomir: No, agents will not read it. Only `AGENTS.md` in the root is read.

AGENTS.md
randomir: This line doesn't mandate one testing framework over another, quite the contrary. This command is **repo-specific** and it instructs agents on how to run tests in this repo, without requiring them to guess.

AGENTS.md
randomir: This command only gives specific instruction on **how** to create a new venv -- if the agent/user asks for that. It's repo specific.

I run Claude in a sandbox, and don't care if it creates a temporary venv for testing.

.agents/AGENTS.baseline.md
randomir: Much more useful and efficient is to have the exact command the agent needs to run in order to setup env for dev, testing, etc...

.agents/AGENTS.baseline.md
randomir: Instructions should be as short and clear as possible. IMO, your suggestion is already covered with other working style bullets.

.agents/AGENTS.baseline.md
randomir: I went back and forth on this one, but Claude says it's useful. I'm adding it.

.agents/AGENTS.baseline.md
randomir: Already implied by "before finishing". The feature/fix is obviously not finished if tests fail!

Generally, a new instruction should be added only if you need to correct or instruct the agent a couple of times. I never had to instruct Claude: "ok, now fix what you broke".

.agents/AGENTS.baseline.md
randomir: Makes sense as some fixes might get skipped due to "make the smallest change".

.agents/AGENTS.template.md
randomir: I think it does what's expected. A feature branch contains the PR changes.

randomir:
> Minor, but can we move the agent files into an `agents` or `.agents` folder and still keep them discoverable? I'm not fond of having too many added files in root.

Yes, I was considering that as well. I'll move the baseline and template files to `.agents/`, and leave the rest in the root. Additionally, we could move `CLAUDE.md` to `.claude/`, but until we have settings or other claude-specific stuff to put in there, `-file,+dir` saves us nothing.

.agents/AGENTS.template.md
qci-amos: How is it controlled which files agents will read vs not? In my experience they read everything they can find.

AGENTS.md
qci-amos: Ok, so these agent files are not intended to be copy/pasted unedited across the other Ocean repos? Do you have a process in mind for centralizing and distributing agent files?

.agents/AGENTS.baseline.md
qci-amos: fwiw, I don't think I've noticed an agent ever doing this. Instead, there's a bit of an opposite problem: it takes the user's prompt too seriously sometimes. Furthermore, in the wording here, "the problem" is implicitly determined by the "ask", so there's a bit of circularity.

Claude suggests this rewording:
> Only edit the files needed to fix the described problem. Don't rename, reformat, extract, or add error handling that isn't required for the fix. If you spot unrelated bugs or cleanup worth doing, list them at the end instead of doing them.

.agents/AGENTS.template.md
randomir: Allegedly, all agents except claude *read* [`AGENTS.md`](https://agents.md/), in addition to their vendor-specific files, like `CLAUDE.md`/`.claude/*`, `.github/copilot-instructions.md`, etc.

When I say *read*, I mean read with intent to find configuration and instructions.

AGENTS.md
randomir: No. I guess the PR description/release note is not as clear as I hoped for.

The core rules are in this repo only, under `AGENTS.baseline.md`.

Each repo will have an instance of `AGENTS.template.md`, specialized for that project (for Ocean SDK repo, that's `AGENTS.md`). In those project rules we instruct the agent to read the baseline rules from the public Ocean repo.

.agents/AGENTS.baseline.md
qci-amos: That's a pretty abstract requirement so I'm not sure how an AI would interpret/handle it.... 

This rationale:

> One of Guido’s key insights is that code is read much more often than it is written.

specifically seems it'll be obsolete soon, if not already in some contexts. Furthermore, the more idiosyncratic a "style" is, the harder it is to read, not easier, for both humans and AI. I'm worried that a constraint like this would impact the quality AI's output.

I generally find that prompts find better results when I minimize the constraints in my prompt...

.agents/AGENTS.baseline.md
qci-amos: It might be worth an explicit request to it not to automatically commit/push/pull.

AGENTS.md
qci-amos: So what does a guideline like this improve in an agents file? What you've written is either implicitly understood or unnecessarily constraining to AI.

.agents/AGENTS.baseline.md
randomir: I'm sure agents would (or soon will) prefer to output machine code in binary.

We're adding these guidelines in the hope our open source software will still be consumed/read, used, and contributed to by humans. So, yes, I believe we want to constrain LLM's output to fit us, not the other way around. At least not yet. :smile:

.agents/AGENTS.baseline.md
randomir: I've noticed Claude vibe-coding the shit out of a simple request, adding abstraction layers and over-engineering. Just like a clueless junior developer. Unless specifically asked not to do it. :shrug: 

But as models and agents evolve, rules such as these might get absorbed and implicitly followed.. we'll need to regularly update our rules, that's a given.

AGENTS.md
randomir: You mean, the command? It gives a clear and explicit recipe for the agent, so it doesn't waste time on inferring the stuff that's well defined in the team's/project's (human) context.

.agents/AGENTS.baseline.md
thisac: I would also lean towards keeping to the _readable_ code standards that we have for now.

.agents/AGENTS.baseline.md
thisac: > It might be worth an explicit request to it not to automatically commit/push/pull.

I think adding something like that could be good also, since I definitely do not want it to do those things without my prior explicit approval (which I would likely not give).

AGENTS.md
thisac: I like having this command here, and I understand the use-case, but I've noticed that it can sometimes create a venv without my informed approval, so having it have to be explicitly approved would IMO be better.

.agents/AGENTS.baseline.md
thisac: Fair enough. If it seems to work.

.agents/AGENTS.template.md
thisac: I reckon much work will be done without a PR being up, so I'm not sure this is correct (although it might still get the gist, so to say). Since we don't want it to care about commit/push/pull, etc. I'd say it shouldn't be concerned with PR, or likely branch, either.

.agents/AGENTS.baseline.md
randomir: IMO, personal preferences like these (branch or not, commit or not, push or not) should be encoded in a local, user-specific file, not here.

.agents/AGENTS.baseline.md
qci-amos: I've certainly seen that it's verbose... but generally I find that it takes the user's prompt as the over-arching source of truth for what it should do. Therefore, the problem you're describing may be the prompter's fault more-so than AI. Correspondingly, maybe what you really want the agent file to do is something like "prompt the user if the change seems bigger than they might be expecting"?

.agents/AGENTS.baseline.md
qci-amos: First, a note: I'm not familiar with Ocean conventions so I don't know what is really meant by "local consistency". I see this page:
https://docs.dwavequantum.com/en/latest/ocean/contribute.html#submitting-changes
which is more explicit about expectations, but is also not worded as if there are package-level exceptions, so I'm not sure what the gap is (nor which such gaps are worth insisting on preserving).

My concern is that what you might be indicating here is a request to be locally-consistent over org and/or industry-consistent. I don't think that is stable or well-defined and it may result in a compromise that isn't optimal for either human nor ai.

What I'm getting at is how are you defining "readable" and "style" and I think some standards like `ruff` formatting, requiring type annotations as much as possible, rigorously maintained docstrings. I think things like that are now low hanging fruit to deliver, and benefit both machine and human. Formatting is really the biggest single item, but I view that as a solved problem these days. Yes, there's a one time cost of reformatting pre-existing code, but I view that to be well worth it in the long run (and there's a way you can ignore those changesets in git blame). My sense is all this is a big part of why OpenAI acquired Astral.

If what you mean are things like variable naming conventions, which probably benefit both, then I think you should spell out some requirements since in general these are fairly arbitrary but (as long as the names aren't stale) can dramatically improve legibility.

Personally, I've long appreciated the Google style guide:
https://google.github.io/styleguide/pyguide.html

and I've appealed to it in past discussions when I've needed a tie breaker during PR review.

.agents/AGENTS.template.md
qci-amos: but what you're suggesting is that it won't read `AGENTS.template.md` even though it's in a folder called `.agents` and uses imperative wording?

One suggestion I've made elsewhere is that we start a NEW repo somewhere with AI recommendations, skills, and best practices that people may adapt as they like a la carte

.agents/AGENTS.baseline.md
randomir: Exactly. We use/recommend the use of google styleguide as well. Check the [final section](https://google.github.io/styleguide/pyguide.html#4-parting-words) please:

> BE CONSISTENT.
>
> If you’re editing code, take a few minutes to look at the code around you and determine its style. If they use _idx suffixes in index variable names, you should too. If their comments have little boxes of hash marks around them, make your comments have little boxes of hash marks around them too.
>
> [...]

.agents/AGENTS.template.md
randomir: Not sure all agents follow the same workflow, but Claude has only a handful of hard-coded locations it will look for the instructions file (system/managed policy, user instructions, user rules, project instructions in `CLAUDE.md`, project rules in `.claude/rules/` and local instructions `CLAUDE.local.md`). These are read before anything else. It won't randomly read every file in the repo (and waste context), unless it's instructed to do so. What's read afterwards is treated as data, not instructions (by design). In practice there are no guarantees, and if the template (or whatever other) file is read, yes it might be interpreted as instructions -- but that's an edge/attack case, not the default/standard behavior.

PR #488, opened 2026-09-10
Add reference to GM

JoelPasvolsky:
#### What does this implement/fix?
Adds a front-page reference to GM package

#### Additional information

- Merge only in the next SDK release
- Thank you, @k8culver for the GM icon!

#### AI Generation Disclosure
No AI tools used

image

CC @randomir, @arcondello, @qwriter, @qci-amos, @qci-petrenko 
Minimal to start with.


thisac (approved):

docs/_images/gate_model_icon.svg
qci-amos, line 1: I'm good with this example, but it's also worth considering using basis gates `sx` and `cz` instead of `h` and `cx`

